PR #62, opened 2020-08-31
fix(general): clean up typos and code

michaelbrewer:
**Issue #, if available:**

## Description of changes:

Changes:
* Fix some of the markdown used
* Removed unused imports
* LambdaHandlerProcessor - Make null check easier to read
* LambdaJsonLayout - Remove unused objectMapper
* Fix typos

**Checklist**

 

* [X] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-java/#tenets)
* [X] Update tests
* [X] Update docs
* [X] PR title follows [conventional commit semantics]()

### DIFF
--- a/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
+++ b/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
@@ -13,18 +13,22 @@
  */
 package software.amazon.lambda.powertools.core.internal;
 
-import java.io.InputStream;
-import java.io.OutputStream;
-
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.RequestStreamHandler;
 import org.aspectj.lang.ProceedingJoinPoint;
 
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Optional;
+
 public final class LambdaHandlerProcessor {
-    private static String SERVICE_NAME = System.getenv("POWERTOOLS_SERVICE_NAME") != null
-            ? System.getenv("POWERTOOLS_SERVICE_NAME") : "service_undefined";
+    private static String SERVICE_NAME = Optional.ofNullable(System.getenv("POWERTOOLS_SERVICE_NAME")).orElse("service_undefined");
     private static Boolean IS_COLD_START = null;
+
+    private LambdaHandlerProcessor() {
+        // Hide default constructor
+    }
 
     public static boolean isHandlerMethod(ProceedingJoinPoint pjp) {
         return "handleRequest".equals(pjp.getSignature().getName());
--- a/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
+++ b/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
@@ -20,27 +20,27 @@ import org.aspectj.lang.ProceedingJoinPoint;
 
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.Optional;
 
 public final class LambdaHandlerProcessor {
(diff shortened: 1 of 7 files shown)


msailes:
Thanks for the PR Michael, just looking at this now.

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
```diff
@@ -22,7 +22,7 @@
 import org.aspectj.lang.ProceedingJoinPoint;
 
 public final class LambdaHandlerProcessor {
-    private static String SERVICE_NAME = null != System.getenv("POWERTOOLS_SERVICE_NAME")
+    private static String SERVICE_NAME = System.getenv("POWERTOOLS_SERVICE_NAME") != null
```
pankajagrawal16, line 25: In order to keep it consistent with how we will like to check for null, I will prefer if we keep constant(null) to the left.

pankajagrawal16 (commented):
@michaelbrewer  Thanks for updates to some of the docs and typos. Muck appreciated:) Just one suggestion.

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
michaelbrewer: Ok sure, i can revert this, just confusing to read left to right.

Maybe this would be cleaner and work with java8

```java
private static String SERVICE_NAME = Optional.ofNullable(System.getenv("POWERTOOLS_SERVICE_NAME")).orElse("service_undefined")
```

michaelbrewer: @pankajagrawal16 ☝️

michaelbrewer: @pankajagrawal16 `SERVICE_NAME` and `IS_COLD_START` look like constants, but are actually mutable due to the unit tests requiring this to be mutable. Maybe we should change either the unit tests or these misleading names.

michaelbrewer:
@pankajagrawal16 - maybe where possible we make things `final` as it reduces any kind of side effects. I just did not want to make too many trivial changes. 

Maybe there anti-bike shedding type of autoformatter do java like "black" for python or "ktlint" for kotlin. To clean up imports and naming conversions.

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
pankajagrawal16: > Ok sure, i can revert this, just confusing to read left to right.
> 
> 
> 
> Maybe this would be cleaner and work with java8
> 
> 
> 
> ```java
> 
> private static String SERVICE_NAME = Optional.ofNullable(System.getenv("POWERTOOLS_SERVICE_NAME")).orElse("service_undefined")
> 
> ```

We have intentionally avoided use of optional here considering its a simple null check which belongs to an internal class.

I am a big fan of functional interfaces in java myself and have to always find a balance when I would like to use it.

Prefer using optional on public API that is public to let users ability to write functional code and not worry about nulls.

But these are just personal preferences too. Considering this being a internal class, I will still be happy to leave it the way it was. Thanks for the input. 🙂

pankajagrawal16:
> @pankajagrawal16 - maybe where possible we make things `final` as it reduces any kind of side effects. I just did not want to make too many trivial changes. 
> 
> Maybe there anti-bike shedding type of autoformatter do java like "black" for python or "ktlint" for kotlin. To clean up imports and naming conversions. 

Having autoformatter sounds like a good idea, but I will suggest please open a RFC for it so that we can track it. As far as client of the library is concerned, it doesn't impact them directly, but definitely worth evaluating at some point.

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
michaelbrewer: @pankajagrawal16  - ok i reverted back to the original code.

pankajagrawal16 (approved):
Looks good 👍

PR #90, opened 2020-09-08
powertools-parameters module

jeromevdl:
## Description of changes:

Add parameters module, to easily retrieve parameters from SSM Parameter Store and Secrets Manager, put them in cache for further reuse.

**Checklist**

* [x] Update tests
* [x] Update docs => javadoc only, will add the doc if we are agree on the code

## Breaking change checklist

New module, no impact on other ones, but probably need to increase version accordingly.


pankajagrawal16:
Thanks @jeromevdl for the contribution. I will have a look at the PR.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SSMProviderTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.openMocks;
+
+public class SSMProviderTest {
```
pankajagrawal16, line 33: These test are again testing on a default method `getValue`which is still ok for unit testing, I would very much like to see end to end tests on these providers . Something like Having a instance from `ParamManagr` with a mocked sdk client and then validating different interactions etc based on cache value.

pankajagrawal16, line 33: Also try using assertj-core for fluent assertions. It really simplifies and makes test assertion much more human readable and easy to understand.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SecretsProviderTest.java
```diff
@@ -0,0 +1,74 @@
...
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest;
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
+
+import java.util.Base64;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.mockito.MockitoAnnotations.openMocks;
+
+public class SecretsProviderTest {
```
pankajagrawal16, line 33: Some of the  same suggestions as for `SSMProviderTest `

pankajagrawal16 (commented):
First of all, thanks for this contribution. It will be a great utility to have.

I have few suggestions/questions which I have tried to explain. Let me know your thoughts and we can discuss further.

Regarding docs, I do agree, once we have the code change finalized, then start with documenting it.(Avoids rework)

powertools-parameters/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>ssm</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>secretsmanager</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache-client</artifactId>
```
msailes, line 58: Can we try this new release?

https://aws.amazon.com/blogs/developer/introducing-aws-common-runtime-http-client-in-the-aws-sdk-for-java-2-x/

powertools-parameters/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.aspectj</groupId>
+            <artifactId>aspectjrt</artifactId>
+            <scope>compile</scope>
+        </dependency>
+
+        <!-- Test dependencies -->
```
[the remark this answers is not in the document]
jeromevdl: my bad, didn't see that. Will replace.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,195 @@
...
+        String value = getValue(key);
+
+        String transformedValue = value;
+        if (transformerClass != null) {
+            if (!BasicTransformer.class.isAssignableFrom(transformerClass)) {
+                resetToDefaults();
+                throw new IllegalArgumentException("Wrong Transformer for a String, choose a BasicTransformer");
+            }
+            try {
+                BasicTransformer transformer = (BasicTransformer) transformerClass.newInstance();
```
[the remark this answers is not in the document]
jeromevdl: replaced with getDeclaredConstructor().newInstance(null);

msailes:
Could we get an update to an existing example, or a new example to show the developer experience? Thanks.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+        T transformedValue = transformer.applyTransformation(value, targetClass);
+
+        store.put(key, transformedValue, Instant.now().plus(maxAge, ChronoUnit.SECONDS).toEpochMilli());
+
+        resetToDefaults();
+        return transformedValue;
+    }
+
+    boolean hasNotExpired(String key) {
```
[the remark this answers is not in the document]
jeromevdl: Not sure how to achieve this, as I explicitly want to be sure the data has not expired. I won't know just with the get, as this is the objective of the lib, to hide the underlying complexity of the cache.

pom.xml
```diff
@@ -93,6 +94,13 @@
                 <artifactId>aws-lambda-java-events</artifactId>
                 <version>${lambda.events.version}</version>
             </dependency>
+            <dependency>
```
msailes, line 97: By adding this, can we remove anything else?

powertools-parameters/pom.xml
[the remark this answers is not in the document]
pankajagrawal16: You wont need to replace anything, you need junit deps anyways .. but just about bring inn assertj as a test dependency and use for fluent assertions.

powertools-parameters/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+        <groupId>software.amazon.lambda</groupId>
+        <version>0.2.0-beta</version>
+    </parent>
+
+    <artifactId>powertools-parameters</artifactId>
+
+    <name>AWS Lambda Powertools Java library Parameters</name>
+
+    <description>
+        Set of utilities to retrieve parameters from Secrets Manager, SSM Parameter Store or DynamoDB
```
msailes, line 18: ✅

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Get the value of a parameter, either from the underlying store or a cached value (if not expired).<br/>
+     * Using this method, you must apply a transformation (eg. json/xml to Object). <br/>
+     * Set a {@link Transformer} with {@link #withTransformation(Class)}.<br/><br/>
+     * If you need a simpler transformation (to String), use {@link #get(String)} method instead of this one.
+     *
+     * @param key         key of the parameter
+     * @param targetClass class of the target Object (after transformation)
+     * @return the Object (T) value of the parameter
```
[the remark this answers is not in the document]
jeromevdl: will add that.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/transform/JsonTransformer.java
```diff
@@ -0,0 +1,37 @@
...
+import software.amazon.lambda.powertools.parameters.exception.TransformationException;
+
+/**
+ * Transformer that transform a json string into an Object. Based on Jackson.
+ *
+ * @param <T> type of the Object to create during transformation.
+ */
+public class JsonTransformer<T> implements Transformer<T> {
+
+    private ObjectMapper mapper = new ObjectMapper();
```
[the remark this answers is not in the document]
jeromevdl: sure

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,195 @@
...
+        this.transformerClass = transformerClass;
+        return (T) this;
+    }
+
+    /**
+     * Abstract method to tell to retrieve parameters recursively. Only available for {@link SSMProvider}.
+     * @param <T>
+     * @return
+     */
+    public <T extends BaseProvider> T recursive() {
```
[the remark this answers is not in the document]
jeromevdl: recursive is only available for SSM when getting multiple parameters. I don't want to force to implement a useless method in other child class (and do what: throw a not implemented exception ?)

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,195 @@
...
+        return (T) this;
+    }
+
+    /**
+     * Abstract method to tell to decrypt parameters. Only available for {@link SSMProvider}.
+     * @return
+     */
+    public <T extends BaseProvider> T withDecryption() {
```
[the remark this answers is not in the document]
jeromevdl: decryption is only available for SSM. I don't want to force to implement a useless method in other child class (and do what: throw a not implemented exception ?)

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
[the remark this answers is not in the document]
pankajagrawal16: I would end up using `java.time.Clock` to control how instant behaves and based on that validate the behavior.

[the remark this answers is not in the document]
pankajagrawal16: and would make sense to test as part of DataStore instead.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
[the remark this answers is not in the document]
pankajagrawal16: Yes, I would have preferred that instead. In most of the cases, users will be using the providers that are built in the library like SSM etc and thats probably one of the reason they would want to use this utility to avoid creating one themselves.

On the other hand if this is SSM specific method, do we even need it on BaseProvider then ?

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+import java.util.Base64;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+import static org.mockito.MockitoAnnotations.openMocks;
+import static software.amazon.lambda.powertools.parameters.transform.Transformer.base64;
+import static software.amazon.lambda.powertools.parameters.transform.Transformer.json;
+
+public class BaseProviderTest {
```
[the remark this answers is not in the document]
jeromevdl: As we have a cache underneath, with TTL, I'm not sure how you can actually verify that it works if you don't check it expired or not. 
Also regarding the Spy, this is simply to verify that the underlying provider (the implementation) actually retrieves data from the store.
I'm open for suggestions as I don't know what to change.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,195 @@
...
+     * {@link software.amazon.lambda.powertools.parameters.transform.Base64Transformer} and {@link software.amazon.lambda.powertools.parameters.transform.JsonTransformer}
+     * are provided for respectively base64 and json content. You can also write your own (see {@link Transformer}).
+     *
+     * <b>Not Thread Safe</b>: calling this method simultaneously by several threads
+     * can lead to errors (one Transformer for the wrong target type)<br/>
+     *
+     * @param transformerClass Class of the transformer to apply. For convenience, you can use {@link Transformer#json} or {@link Transformer#base64} shortcuts.
+     * @return the provider itself in order to chain calls (eg. <code>provider.withTransformation(json).get("key", MyObject.class)</code>).
+     */
+    public <T extends BaseProvider> T withTransformation(Class<? extends Transformer> transformerClass) {
```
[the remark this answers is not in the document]
jeromevdl: Transformer is not mandatory. Most of the time, you'll just retrieve a String, thus you don't need a Transformer.
And also, you generally don't have a unique type for all your parameters. Not sure that make sense to have a common Transformer.
Finally, I did almost the same as in the python version (bit more complex in java as we need the Type).

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/ParamManager.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private static SecretsProvider secretsProvider;
+    private static SSMProvider ssmProvider;
+
+    /**
+     * Get a {@link SecretsProvider} with default {@link SecretsManagerClient}.<br/>
+     * If you need to customize the region, or other part of the client, use {@link ParamManager#getSecretsProvider(SecretsManagerClient)} instead.
+     * @return a {@link SecretsProvider}
+     */
+    public static SecretsProvider getSecretsProvider() {
```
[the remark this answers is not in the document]
jeromevdl: Yes, I asked myself about this one too, and finally decided to do like Log4J:
`Logger LOG = LogManager.getLogger(...)`

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/ParamManager.java
```diff
@@ -0,0 +1,75 @@
...
+package software.amazon.lambda.powertools.parameters;
+
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
+import software.amazon.awssdk.services.ssm.SsmClient;
+
+/**
+ * Utility class to retrieve instances of parameter providers.
+ * Each instance is unique (singleton).
+ */
+public final class ParamManager {
```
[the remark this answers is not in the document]
jeromevdl: Probably I can add few ones. Not a lot of value in them but to have a security belt OK.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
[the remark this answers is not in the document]
pankajagrawal16: Sure, I can make a suggestion and we can discuss if that looks better.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SSMProviderTest.java
jeromevdl: getValue is actually what we want to test in those classes. This is the core method that retrieves the parameter with a mocked client, exactly what you mention. Not sure an integration test which does the same has any value. Maybe a real integration test with a real call to SSM would be interesting.

jeromevdl: Will do

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SecretsProviderTest.java
jeromevdl: Same answer, we must favor unit tests over integration tests (test pyramid). This class tests exactly what she needs to, retrieve the parameter in the 'real' store (which is mocked). If we want integration test, let's do one with a real call a Secrets Manager (maybe with localstack).

powertools-parameters/pom.xml
jeromevdl: This is quite cutting edge ;) Maybe too much, as it is only a PREVIEW...

msailes: Never mind, I think this is just for async

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SecretsProviderTest.java
pankajagrawal16: yes i think we should have one full stack stack showcasing all the public api method calls. As long as we inject mocked client to the provider, we can proceed cover it pretty well

pom.xml
jeromevdl: I'm not sure as the sdk was actually not used in other modules (xray uses different dependencies, logging does not use it at all). It's a bom, so just a way to specify dependencyManagement for all the sdk dependencies.

powertools-parameters/pom.xml
jeromevdl: should remove DynamoDB a it is not yet implemented...

powertools-parameters/pom.xml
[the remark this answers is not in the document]
jeromevdl: I mean replace assertEquals with assertThat...isEqual and all the rest.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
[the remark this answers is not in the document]
pankajagrawal16: So basically my suggestion is to test DataStore class directly to validate the functionality of `hasNotExpired` method instead of creating a default method inside BaseProvider which is just used for testing.

Test BaseProvider just for validating fetch from datastore/cache or from remote behavior. And that would simplify test alot.

When it comes to validating without thread.sleep for expiration, if we leverage java.time.clock to manipulate the time, we check the expiration behavior.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/ParamManager.java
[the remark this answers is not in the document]
pankajagrawal16: I actually think we can use this class to test the utility/provider end to end. So if we inject a mock client in provider and then work on real api methods. But I am ok both ways, if we just unit test this and have more "fullstack" test class separately.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/ParamManager.java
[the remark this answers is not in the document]
pankajagrawal16: Yeah ok. No worries. I guess it just about personal preference.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,195 @@
...
+     * The max age is reset to default (either 5 or a custom value set with {@link #defaultMaxAge}) after each get,
+     * so you need to use this method for each parameter to cache with non-default max age.<br/><br/>
+     *
+     * <b>Not Thread Safe</b>: calling this method simultaneously by several threads
+     * can lead to unwanted cache time for some parameters.<br/>
+     *
+     * @param maxAgeinSec Maximum time (in seconds) to cache the parameter, before calling the underlying parameter store.
+     * @return the provider itself in order to chain calls (eg. <code>provider.withMaxAge(10).get("key")</code>).
+     */
+    public <T extends BaseProvider> T withMaxAge(int maxAgeinSec) {
```
msailes, line 72: Would it be better to support withMaxAge(ChronoUnit unit, int amount)?

I can imagine most parameters would have quite long durations. Over a minute, often over an hour.

pankajagrawal16: Yeah makes sense to me.

jeromevdl: Makes sense, just a bit more long to type...

pankajagrawal16: Yeah we could probably have both ? one with seconds like it is now and other with time unit

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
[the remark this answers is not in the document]
jeromevdl: Not sure to get your preference. For sure, throwing a not implemented exception is a really bad practice (against SOLID principles). Eventually put a log to notify the developer the method is useless in that case.
I've added it to the Base in order to be able to chain all calls. It was only in SSM at the beginning, but then you cannot do provider.recursive().withMaxAge().get() or provider.withMaxAge().recursive().get()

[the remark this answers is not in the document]
pankajagrawal16: aha i see the reason why its there in base class then .  

Hmm may be its ok the way its is rite now then. I dont see any real value in adding log either.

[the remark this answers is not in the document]
pankajagrawal16: I am wondering at this point, what advantage do we get by introducing inheritance here with  BaseProvider ? 

What would the implementation looks like if these were concrete classes on its own ? How much code is duplicated vs complexity it introduces ? 

For clients perspective, if I understand correctly , they still end up calling `ParamManager` static method for getting instance of param provider of their choice, what advantage inheritance gives us here 🤔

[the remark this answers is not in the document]
jeromevdl: The BaseProvider handles the cache mechanism + transformation. It was done just like that in the Python version and it looks good to me. I don't want to duplicate that in several implementations. Moreover if we add DynamoDB after and potentially others. But you are right that it breaks the composition over inheritance pattern... let me think how we could avoid those empty methods and minimize code duplication...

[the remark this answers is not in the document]
pankajagrawal16: Yeah. As long as code duplication is concerned, we can always extract the entire caching logic into a method which can be resued by any provider.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,186 @@
...
+     * If you need a simpler transformation (to String), use {@link #get(String)} method instead of this one.
+     *
+     * @param key         key of the parameter
+     * @param targetClass class of the target Object (after transformation)
+     * @return the Object (T) value of the parameter
+     * @throws IllegalArgumentException if no transformation class was provided through {@link #withTransformation(Class)}
+     * @throws TransformationException  if the transformation could not be done, because of a wrong format or an error during transformation.
+     */
+    @Override
+    public <T> T get(String key, Class<T> targetClass) {
```
pankajagrawal16, line 152: Minor: May be we can wrap try finally logic like this?

```
  private <R> R withResetToDefaults(String key, Function<String, R> function) {
        try {
            return function.apply(key);
        } finally {
            resetToDefaults();
        }
    }

```

And then both the method could just call this wrapper ?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/CacheManager.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class CacheManager {
+    static final Duration DEFAULT_MAX_AGE_SECS = Duration.of(5, SECONDS);
+
+    private final DataStore store;
+    private final NowProvider nowProvider;
+    private Duration defaultMaxAge = DEFAULT_MAX_AGE_SECS;
+    private Duration maxAge = defaultMaxAge;
+
+    public CacheManager(NowProvider nowProvider) {
```
pankajagrawal16, line 29: Instead of having NowProvider can these classes depend on Clock instead ? That is the java8 recommended way of  dealing with manipulation of time

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/DataStore.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Internal store used to cache parameters
+ */
+public class DataStore {
+
+    private final ConcurrentHashMap<String, ValueNode> store;
+    private final NowProvider nowProvider;
+
+    public DataStore(NowProvider nowProvider) {
```
pankajagrawal16, line 27: Instead of having NowProvider can these classes depend on Clock instead ? That is the java8 recommended way of dealing with manipulation of time

pankajagrawal16 (commented):
This has started to look nice. Some minor suggestion. Also if we could have a test where we can see how clients will use these API directly, will be great. Like end to end but just mocking the client and CacheManager with Clock

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/transform/TransformationManagerTest.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    public void notSetTransformer_shouldNotTransform() {
+        assertThat(manager.shouldTransform()).isFalse();
+    }
+
+    @Test
+    public void performBasicTransformation_noTransformer_shouldThrowException() {
+        assertThatIllegalStateException().isThrownBy(() -> manager.performBasicTransformation("value"));
```
pankajagrawal16, line 49: Formatting: I generally prefer formatting these chains to next line. For example: 

```
assertThatIllegalStateException()
                      .isThrownBy(() -> manager.performBasicTransformation("value"));
```

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/transform/JsonTransformerTest.java
```diff
@@ -0,0 +1,52 @@
...
+        assertThat(objectToDeserialize.getBar()).isEqualTo(42);
+        assertThat(objectToDeserialize.getBaz()).isEqualTo(123456789);
+    }
+
+    @Test
+    public void transform_json_shouldTransformInHashMap() throws TransformationException {
+        JsonTransformer<Map> transformation = new JsonTransformer<>();
+
+        Map<String, Object> map = transformation.applyTransformation("{\"foo\":\"Foo\", \"bar\":42, \"baz\":123456789}", Map.class);
+        assertThat(map.get("foo")).isEqualTo("Foo");
```
pankajagrawal16, line 41: You can actually chains these and use containsEntry api

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/cache/CacheManagerTest.java
```diff
@@ -0,0 +1,124 @@
...
+        manager.resetExpirationtime();
+        manager.putInCache("key2", "value2");
+
+        Instant now = Instant.now();
+        Mockito.when(nowProvider.now()).thenReturn(now.plus(40, SECONDS));
+
+        Optional<String> value = manager.getIfNotExpired("key");
+        Optional<String> value2 = manager.getIfNotExpired("key2");
+
+        assertThat(value.isPresent()).isFalse();
```
pankajagrawal16, line 120: You have fluent api on assert to validate its presense.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/cache/CacheManagerTest.java
```diff
@@ -0,0 +1,124 @@
...
+    public void getIfNotExpired_withCustomDefaultExpirationTime_notExpired_shouldReturnValue() {
+        manager.setDefaultExpirationTime(Duration.of(42, SECONDS));
+        manager.putInCache("key", "value");
+
+        Instant now = Instant.now();
+        Mockito.when(nowProvider.now()).thenReturn(now.plus(40, SECONDS));
+
+        Optional<String> value = manager.getIfNotExpired("key");
+
+        assertThat(value.isPresent()).isTrue();
```
pankajagrawal16, line 88: You have fluent api on assert to validate its presence and asset on value directly

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SSMProviderTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        assertThat(params.get("key1")).isEqualTo("foo1");
+        assertThat(params.get("key2")).isEqualTo("foo2");
+        assertThat(params.get("key3")).isEqualTo("foo3");
+
+        List<GetParametersByPathRequest> requestParams = paramByPathCaptor.getAllValues();
+        GetParametersByPathRequest request1 = requestParams.get(0);
+        GetParametersByPathRequest request2 = requestParams.get(1);
+
+        assertThat(request1.path()).isEqualTo("/prod/app1");
```
pankajagrawal16, line 128: You can use all satisfies api on it to validate the assertions that are common.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SSMProviderTest.java
```diff
@@ -0,0 +1,145 @@
...
+    public void getMultiple() {
+        List<Parameter> parameters = new ArrayList<>();
+        parameters.add(Parameter.builder().name("/prod/app1/key1").value("foo1").build());
+        parameters.add(Parameter.builder().name("/prod/app1/key2").value("foo2").build());
+        parameters.add(Parameter.builder().name("/prod/app1/key3").value("foo3").build());
+        GetParametersByPathResponse response = GetParametersByPathResponse.builder().parameters(parameters).build();
+        when(client.getParametersByPath(paramByPathCaptor.capture())).thenReturn(response);
+
+        Map<String, String> params = provider.getMultiple("/prod/app1");
+        assertThat(params.get("key1")).isEqualTo("foo1");
```
pankajagrawal16, line 96: Chain into one assertion

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/CacheManager.java
jeromevdl: OK, I did that. Looks better now.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/DataStore.java
jeromevdl: OK, I did that. Looks better now.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -0,0 +1,186 @@
...
+     * @param key key of the parameter
+     * @return the String value of the parameter
+     * @throws IllegalArgumentException if a wrong transformer class is provided through {@link #withTransformation(Class)}. Needs to be a {@link BasicTransformer}.
+     * @throws TransformationException  if the transformation could not be done, because of a wrong format or an error during transformation.
+     */
+    @Override
+    public String get(String key) {
+        try {
+            Optional<String> valueInCache = cacheManager.getIfNotExpired(key);
+            if (valueInCache.isPresent()) {
```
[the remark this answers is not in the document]
jeromevdl: done, much fluent/clearer

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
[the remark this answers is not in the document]
jeromevdl: Tests are more consistent, but not simpler, as there are quite a lot of cases.
At least, things are more decoupled with the CacheManager and the TransformationManager.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/SSMProviderTest.java
jeromevdl: done

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/transform/TransformationManagerTest.java
jeromevdl: if you prefer.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
jeromevdl: 2 different signatures, don't think it's worth it.

pankajagrawal16: Ok no problem.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/ParamManager.java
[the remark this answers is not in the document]
jeromevdl: see [ParamManagerIntegrationTest.java](powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/ParamManagerIntegrationTest.java)

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/ParamManagerIntegrationTest.java
```diff
@@ -0,0 +1,127 @@
...
+    @Captor
+    ArgumentCaptor<GetSecretValueRequest> secretsCaptor;
+
+
+    @BeforeEach
+    public void setup() throws NoSuchFieldException, IllegalAccessException {
+        openMocks(this);
+
+        // A bit dirty but necessary as everything is static in ParamManager
+        Field ssmProviderField = ParamManager.class.getDeclaredField("ssmProvider");
```
pankajagrawal16, line 61: We have used commons lang  in test context in other module to use Apache reflection utils. If needed you might check that too.

powertools-parameters/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
```
pankajagrawal16, line 92: Is this needed ? 🤔

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/NowProvider.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package software.amazon.lambda.powertools.parameters.cache;
+
+import java.time.Instant;
+
+public class NowProvider {
```
pankajagrawal16, line 18: This looks like being unused now ?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/transform/TransformationManager.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param value the value to transform
+     * @param targetClass the type of the target object.
+     * @return the value transformed in an object ot type T.
+     */
+    public <T> T performComplexTransformation(String value, Class<T> targetClass) {
+        if (transformer == null) {
+            throw new IllegalStateException("You cannot perform a transformation without Transformer, use the provider.withTransformation() method to specify it.");
+        }
+
+        Transformer<T> complexTransformer;
```
pankajagrawal16, line 78: Inline within try ?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/NowProvider.java
jeromevdl: sure, I remove

powertools-parameters/pom.xml
jeromevdl: no it isn't, removed

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/transform/TransformationManager.java
jeromevdl: I'm not sure for this one... why ?

pankajagrawal16: I just menat something like this :

```
Transformer<T> complexTransformer = transformer.getDeclaredConstructor().newInstance(null);
```
Instead of declaring it outside try

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/ParamManagerIntegrationTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Mock
+    SecretsManagerClient secretsManagerClient;
+
+    @Captor
+    ArgumentCaptor<GetSecretValueRequest> secretsCaptor;
+
+
+    @BeforeEach
+    public void setup() throws NoSuchFieldException, IllegalAccessException {
```
pankajagrawal16, line 57: `NoSuchFieldException` is not needed ?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/CacheManager.java
```diff
@@ -0,0 +1,57 @@
...
+    public void setDefaultExpirationTime(Duration duration) {
+        this.defaultMaxAge = duration;
+        this.maxAge = duration;
+    }
+
+    public <T> void putInCache(String key, T value) {
+        store.put(key, value, Clock.systemDefaultZone().instant().plus(maxAge));
+    }
+
+    public void resetExpirationtime() {
```
pankajagrawal16, line 54: Nitpicking 🙈  But may be `resetExpirationTime ` ?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/transform/TransformationManager.java
jeromevdl: Ok

pankajagrawal16 (approved):
This looks really nice to me with all the test etc. 

We should make sure that we have the docs updated as well to document the utility, either as part of this PR or a separate is ok too.

pankajagrawal16 (dismissed):

pankajagrawal16 (approved):

jeromevdl:
Will create the doc and a sample once this PR is validated.

msailes:
> Will create the doc and a sample once this PR is validated.

Yeah, lets move onto writing the docs and sample!

pankajagrawal16 (approved):
LGTM 🚀

Issue #58, opened 2020-08-31
Gradle plug-in for compile-time weave (CTW) 


michaelbrewer:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Please include a gradle plugin for setting up the compile-time weave (CTW). I think this would be a far more popular option than maven. 

**Describe alternatives you've considered**

**Additional context**

msailes:
Hi Michael,

We'll have a look at this, thanks.

Mark

michaelbrewer:
@msailes thanks once we have this. It will be easy to incorporate this into the majority of java lambda projects using gradle.

pankajagrawal16:
Closed by #147

michaelbrewer:
Thanks @pankajagrawal16 !

PR #99, opened 2020-09-22
chore: add example for gradle build 

am29d:
**Issue #, if available:** #58 

## Description of changes: 

Add gradle build example that uses aspectj plugin and compile-time-weave feature. I have merged both configuration into one project for easier maintenance.

**Checklist**

 

* [ ] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:


pankajagrawal16 (approved):
Should we update docs page as well ? https://awslabs.github.io/aws-lambda-powertools-java/

am29d:
Good point, I will add a gradle example below maven.

example/template.yaml
```diff
@@ -21,8 +21,6 @@ Resources:
       Environment: # More info about Env Vars: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#environment-object
         Variables:
           POWERTOOLS_SERVICE_NAME: "Payment Service"
-          LOG_LEVEL: INFO
```
pankajagrawal16, line 24: Any specific reason for removing this ?

am29d: no, it was removed due to merge with master to resolve conflicts, I will add it back.

pankajagrawal16 (approved):

example/README.md
```diff
@@ -50,24 +49,24 @@ You can find your API Gateway Endpoint URL in the output values displayed after
 Build your application with the `sam build` command.
 
 ```bash
-sam-app$ sam build
+examples$ sam build
 ```
 
-The SAM CLI installs dependencies defined in `HelloWorldFunction/pom.xml`, creates a deployment package, and saves it in the `.aws-sam/build` folder.
+The SAM CLI installs dependencies defined in `HelloWorldFunction/build.gradle`, creates a deployment package, and saves it in the `.aws-sam/build` folder.
```
msailes, line 55: I think I'd rather have both examples than just gradle.

docs/content/index.mdx
```diff
@@ -66,6 +66,42 @@ And configure the aspectj-maven-plugin to compile-time weave (CTW) the aws-lambd
...
+    implementation 'org.apache.logging.log4j:log4j-api:2.13.3'
+    implementation 'org.apache.logging.log4j:log4j-core:2.13.3'
+
+    testImplementation 'junit:junit:4.12'
+}
+
+
+```
+
+**Note:** keep in mind to add `aspectjVersion = '1.9.6'` to the `gradle.properties` file. The aspectj plugin works at the moment only with gradle 6.x. There is an [open PR with a fix](https://github.com/arendd/AspectjGradlePlugin/pull/5), once merged the same build configuraiton can be used with gradle 5.x.
```
orozcoadrian, line 103:
```suggestion
**Note:** keep in mind to add `aspectjVersion = '1.9.6'` to the `gradle.properties` file. The aspectj plugin works at the moment only with gradle 6.x. There is an [open PR with a fix](https://github.com/arendd/AspectjGradlePlugin/pull/5), once merged the same build configuration can be used with gradle 5.x.
```

example/README.md
```diff
@@ -50,24 +49,24 @@ You can find your API Gateway Endpoint URL in the output values displayed after
 Build your application with the `sam build` command.
 
 ```bash
-sam-app$ sam build
+examples$ sam build
```
orozcoadrian, line 52:
```suggestion
example$ sam build
```

example/README.md
am29d: Initially I had two separate projects, but because only the build configuration was the difference it makes sense to pull them together in one project and have two options `build.gradle` and `pom.xml`. I will add the documentation to reflect that the user have a choice.

am29d: Interesting, `sam build` favors gradle over maven, if both configurations are in the same root folder 😄

pankajagrawal16: haha .. good to know that :D

am29d:
Hi there, I have to retract the PR, due to the current AspectJ Gradle plugin issues. The plugin we were planning to use [does not work with Gradle 5.x](https://github.com/arendd/AspectjGradlePlugin/pull/5), there is an open PR. Besides the plugin was build with Java12 which causes RuntimeExceptions once deployed with Java11 runtime.  

I need more time to research the AspectJ Gradle plugins to find a working solution for the current Lambda Java runtime configurations and both Gradle 5.x and Gradle 6.x version. Sorry for dragging this out, I know lots of folks are using Gradle and are looking into how to use the powertools in their setup.

orozcoadrian:
fwiw, I'm using the logging module (no xray, etc) with the following:
- gradle-6.6.1
- java11
- i'm NOT using any aspectj dependencies nor configurations

am29d:
thx for the info @orozcoadrian, I assume you use load time weaving mechanisms which increases start time, compared to compile time weave. I dont have the number how much overhead it produces, but I assume it would be more efficient to weave during compile time 🤔 . I am not very familiar with all the AspectJ weaving option yet, but [this explanations](https://stackoverflow.com/questions/39448543/run-time-aop-vs-compile-time-aop/39556178#39556178) helped to wrap my head around pros and cons. I wonder what it would mean for the Lambda runtime.

PR #123, opened 2020-10-06
fix: Make slf4j work along with log4j2

pankajagrawal16:
**Issue #, if available:**

## Description of changes:

Fixes warning like below:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
-- | --
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,6 +125,11 @@
             </dependency>
             <dependency>
                 <groupId>org.apache.logging.log4j</groupId>
+                <artifactId>log4j-slf4j-impl</artifactId>
+                <version>${log4j.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.logging.log4j</groupId>
                 <artifactId>log4j-api</artifactId>
                 <version>${log4j.version}</version>
             </dependency>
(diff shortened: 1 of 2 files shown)


msailes (approved):
approve

jeromevdl (approved):
OK.
 
Just to remember, if SLF4J version is >= 1.8, then we should use `log4j-slf4j18-impl` instead,

see https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/

pankajagrawal16:
> OK.
> 
> Just to remember, if SLF4J version is >= 1.8, then we should use `log4j-slf4j18-impl` instead,
> 
> see https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/

Good point, Currently this is good:

```
[INFO] |  +- software.amazon.awssdk:profiles:jar:2.15.2:compile
[INFO] |  +- software.amazon.awssdk:sdk-core:jar:2.15.2:compile
[INFO] |  |  +- org.slf4j:slf4j-api:jar:1.7.28:compile
[INFO] |  |  \- org.reactivestreams:reactive-streams:jar:1.0.2:compile
```

Issue #297, opened 2021-02-25
[Metrics] Improvements to Powertools Metrics


humanzz:
**Is your feature request related to a problem? Please describe.**
The following requests are based on a recent experience I had using Powerttools Metrics having used https://github.com/awslabs/aws-embedded-metrics-java in different projects before and the workarounds we had to implement and hoping they should be taken care of by Powertools so we can move our older projects to Powertools and have a consistent experience among the different projects.

As such, I'll mention a few points below

**Describe the solution you'd like**
1. **Improve documentation/warnings about the fact that powertools annotations require the Lambda handler method to be called `handlerRequest`**
    - I use CloudGuru Profiler with the setup described in https://docs.aws.amazon.com/codeguru/latest/profiler-ug/lambda-custom.html. This implementation overrides/hides `handleRequest` and instead exposes `requestHandler` method that users should override to add their logic
    - It took me a lot of time debugging/investigating this; time that could've been saved if there was better documentation or warnings about where the annotations are.
    - My workaround was to create a new `BaseHandler` that extends `RequestHandlerWithProfiling`, override `handleRequest` calling `super.handleRequest` and annotating it with the powertools annotations. Then implementing my concrete handlers as subclasses of `BaseHandler`

2. **Metrics emitted by `MetricsUtils.metricsLogger` should have a property for the `requestId`**
    - The Metric EMF entity emitted does not contain a property for the `requestId`. It has other useful things like `traceId`
    - I think the [LambdaMetricsAspect](https://github.com/awslabs/aws-lambda-powertools-java/blob/master/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspect.java) can be updated to add a `requestId` property by default

3. **Metrics emitted by `MetricsUtils.metricsLogger` should allow dimensions to be overriden or unset**
   - By default, they get the following dimensions `"Dimensions":[["LogGroup","ServiceName","ServiceType","Service"]]`.
   - For `ServiceName` and `Service`; that's unnecessary, I believe one to be coming from the EMF Library and the other from Powertools
      - https://github.com/awslabs/aws-embedded-metrics-java/blob/master/src/main/java/software/amazon/cloudwatchlogs/emf/logger/MetricsLogger.java#L174
[...]

PR #305, opened 2021-02-26
feat: single metric utility method to pick default namespace

pankajagrawal16:
**Issue #297 

## Description of changes:

Expose another helper utility method to pick default namespace from top level annotation or env var

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
+++ b/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
@@ -2,7 +2,10 @@ package software.amazon.lambda.powertools.metrics;
 
 import java.util.function.Consumer;
 
+import software.amazon.cloudwatchlogs.emf.config.SystemWrapper;
 import software.amazon.cloudwatchlogs.emf.logger.MetricsLogger;
+import software.amazon.cloudwatchlogs.emf.model.MetricsContext;
+import software.amazon.cloudwatchlogs.emf.model.MetricsLoggerHelper;
 import software.amazon.cloudwatchlogs.emf.model.Unit;
 
 /**
@@ -27,6 +30,29 @@ public final class MetricsUtils {
     }
 
     /**
+     * Add and immediately flush a single metric. It will use the default namespace
+     * specific either on {@link Metrics} annotation or via POWERTOOLS_METRICS_NAMESPACE env var.
+     *
+     * @param name   the name of the metric
+     * @param value  the value of the metric
+     * @param unit   the unit type of the metric
+     * @param logger the MetricsLogger
+     */
+    public static void withSingleMetricOnDefaultNameSpace(final String name,
+                                                          final double value,
+                                                          final Unit unit,
+                                                          final Consumer<MetricsLogger> logger) {
+        MetricsLogger metricsLogger = new MetricsLogger();
+        try {
+            metricsLogger.setNamespace(defaultNameSpace());
+            metricsLogger.putMetric(name, value, unit);
+            logger.accept(metricsLogger);
+        } finally {
+            metricsLogger.flush();
+        }
+    }
+
+    /**
      * Add and immediately flush a single metric.
      *
      * @param name the name of the metric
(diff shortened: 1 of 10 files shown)


powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
```diff
@@ -26,6 +29,29 @@ public static MetricsLogger metricsLogger() {
...
+    /**
+     * Add and immediately flush a single metric. It will use the default namespace
+     * specific either on {@link Metrics} annotation or via POWERTOOLS_METRICS_NAMESPACE env var.
+     *
+     * @param name   the name of the metric
+     * @param value  the value of the metric
+     * @param unit   the unit type of the metric
+     * @param logger the MetricsLogger
+     */
+    public static void withSingleMetricOnDefaultNameSpace(final String name,
```
humanzz, line 41: This is great. Really appreciate that.
I'd say that maybe we can expand this to also set the `AwsRequestId` property. Probably we should do the same for `withSingleMetric`.

If we do the `AwsRequestId` part that will ensure that all metrics `metrcisLogger()` and `withSingleMetric` methods from `MetricsUtils` have the request Id and the traceId which ensures good observability and tracing

I know you've taken a very explicit approach in naming the function, but I think the `OnDefaultNameSpace` suffix can be dropped in favour of documentation calling out the usage of the default namespace.

pankajagrawal16: I kind of like the idea here. Appreciate the feedback.

pankajagrawal16: Suggestion incorporated. Now it will capture request id, trace id whenever possible as property

powertools-metrics/src/test/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspectTest.java
```diff
@@ -71,13 +72,33 @@ void tearDown() {
 
     @Test
     public void metricsWithoutColdStart() {
-        try (MockedStatic<SystemWrapper> mocked = mockStatic(SystemWrapper.class)) {
+        try (MockedStatic<SystemWrapper> mocked = mockStatic(SystemWrapper.class);
+             MockedStatic<software.amazon.lambda.powertools.core.internal.SystemWrapper> internalWrapper = mockStatic(software.amazon.lambda.powertools.core.internal.SystemWrapper.class)) {
```
msailes, line 76: Are these the same class? does it need to be fully qualified?

powertools-metrics/src/test/java/software/amazon/lambda/powertools/metrics/handlers/PowertoolsMetricsEnabledHandler.java
```diff
@@ -16,6 +18,10 @@ public Object handleRequest(Object input, Context context) {
         MetricsLogger metricsLogger = metricsLogger();
         metricsLogger.putMetric("Metric1", 1, Unit.BYTES);
 
+
+        withSingleMetric("Metric2", 1, Unit.COUNT,
```
msailes, line 22: nice!

powertools-metrics/src/test/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspectTest.java
pankajagrawal16: No, not same classes. One is from EMF lib and qualified one from powertools.

msailes (approved):
I think this is a great improvement from some great feedback. 

99% of customers will be happy to see this opinionated action taken. 

Great stuff!

msailes (approved):
Great!

msailes (approved):
👍

PR #308, opened 2021-03-01
chore: migrate docs from gatsby to mkdocs

pankajagrawal16:
**Issue #, if available:**

## Description of changes:

Preview available here https://pankajagrawal16.github.io/aws-lambda-powertools-java/

**Checklist**

 

* [ ] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:


msailes:
I don't think I can review this, bit out of my knowledge range. @heitorlessa maybe?

heitorlessa:
Tip - Add a preview URL from your branch Pankaj at the top of the PR
description —- that way Mark can review its content.

It’s good opportunity to clean up and make navigation easier etc

pankajagrawal16:
Yeah I already added preview link in the description

pankajagrawal16:
https://pankajagrawal16.github.io/aws-lambda-powertools-java/

.github/workflows/docs.yml
```diff
@@ -10,19 +10,19 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - uses: borales/actions-yarn@v2.0.0
+      - uses: actions/checkout@v1
       - name: Set up Python
         uses: actions/setup-python@v1
         with:
           python-version: "3.8"
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12'
```
am29d, line 20: I am not sure, do we use this action here? It is similar in `build-docs.yml`

docs/core/logging.md
```diff
@@ -0,0 +1,223 @@
+---
+title: Logging
```
am29d, line 2: I just noticed small inconsistency within the doc and between python/java docs. In python we use `Logger`, `Tracer`. In Java it is `Logging`, `Tracing`. Down below in the text we switch back to `Logger`. So I would suggest to change it to `Logger` and `Tracer` to keep it consistent.

docs/core/logging.md
```diff
@@ -0,0 +1,223 @@
...
+description: Core utility
+---
+
+Logger provides an opinionated logger with output structured as JSON.
+
+**Key features**
+
+* Capture key fields from Lambda context, cold start and structures logging output as JSON
+* Log Lambda event when instructed (disabled by default)
+    - Enable explicitly via annotation param
```
am29d, line 12: Is the additional level intentional? If yes can we move it up since it only refer to the line before, something like that:
* Log Lambda event when instructed, disabled by default, can be enabled explicitly via annotation param

docs/core/logging.md
```diff
@@ -0,0 +1,223 @@
...
+* Append additional keys to structured log at any point in time
+
+## Initialization
+
+Powertools extends the functionality of Log4J. Below is an example log4j2.xml file, with the LambdaJsonLayout configured.
```
am29d, line 17: I would highlight both the filename `log4j2.xml` and XML attribute `LambdaJsonLayout`, draws the focus of the reader easier if they are scanning the docs. In addition I would highlight the line 5 where the XML attribute is.

docs/core/metrics.md
```diff
@@ -0,0 +1,176 @@
...
+    Generally, this would be an edge case since you [pay for unique metric](https://aws.amazon.com/cloudwatch/pricing/). Keep the following formula in mind:
+    **unique metric = (metric_name + dimension_name + dimension_value)**
+
+=== "App.java"
+
+    ```java
+    withSingleMetric("CustomMetrics2", 1, Unit.COUNT, "Another", (metric) -> {
+        metric.setDimensions(DimensionSet.of("AnotherService", "CustomService"));
+    });
+    ```
```
am29d, line 176: This one is not like the others 😄 . I assume it is within the handler method, but it is not clear where this snipped should be in the code. Having it in the scope of the handler would help to understand where it can/should be used.

docs/index.md
```diff
@@ -0,0 +1,145 @@
...
+                     <execution>
+                         <goals>
+                             <goal>compile</goal>
+                         </goals>
+                     </execution>
+                 </executions>
+            </plugin>
+            ...
+        </plugins>
+    </build>
```
am29d, line 104: I think we need both parts of the pom to be present right? So I would put it into one file to reduce clicks and easier copy paste for the users.

docs/utilities/batch.md
```diff
@@ -0,0 +1,258 @@
...
+* A simple interface for individually processing messages from a batch
+
+**Background**
+
+When using SQS as a Lambda event source mapping, Lambda functions can be triggered with a batch of messages from SQS.
+
+If your function fails to process any message from the batch, the entire batch returns to your SQS queue, and your Lambda function will be triggered with the same batch again.
+
+With this utility, messages within a batch will be handled individually - only messages that were not successfully processed
+are returned to the queue.
```
am29d, line 20: This can be one paragraph, the break between each sentence suggest they might be disconnected.

docs/utilities/batch.md
```diff
@@ -0,0 +1,258 @@
...
+                         <goals>
+                             <goal>compile</goal>
+                         </goals>
+                     </execution>
+                 </executions>
+            </plugin>
+            ...
+        </plugins>
+    </build>
+    ```
```
am29d, line 75: Similar to "Homepage" page, this can be in one tab, since we need the content from both tabs to configure the pom.

docs/utilities/batch.md
```diff
@@ -0,0 +1,258 @@
...
+    - **2)** Raise `SQSBatchProcessingException` to ensure failed messages return to your SQS queue
+
+The only difference is that **SqsUtils Utility API** will give you access to return from the processed messages if you need. Exception `SQSBatchProcessingException` thrown from the
+utility will have access to both successful and failed messaged along with failure exceptions.
+
+## Functional Interface SqsMessageHandler
+
+Both [annotation](#sqsbatch-annotation) and [SqsUtils Utility API](#sqsutils-utility-api) requires an implementation of functional interface `SqsMessageHandler`.
+
+This implementation is responsible for processing each individual message from the batch, and to raise an exception if unable to process any of the messages sent.
```
am29d, line 109: Can be one paragraph.

am29d (commented):
There is one general thought I had when going through the docs. Many code examples don't contain the import statements. While we don't have to bloat the import section of the examples, it is helpful to mention at least the imports from the powertools library. Sometimes IDE can't autoimport, sometimes the class names might be similar within other packages. I had this experience few times when reading other docs and was frustrated to understand where specific classes are coming from.

.github/workflows/docs.yml
pankajagrawal16: I guess its just copy pasta from python version. I will check and remove if not needed

docs/core/logging.md
pankajagrawal16: I will check this, I guess in Java its better to keep calling them `Logging` and `Tracing` since thats what annotations are called. I will check for places in docs where its used otherwise

pankajagrawal16:
@am29d Thanks for great suggestions. I believe I have all of them in place now. Give it another shot. Live preview is updated as well https://me.pankaagr.cloud/aws-lambda-powertools-java/

am29d (approved):
LGTM

PR #330, opened 2021-03-15
feat: remove deprecated attributes on Tracing annotation

pankajagrawal16:
**Issue #, if available:**

## Description of changes:

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:


msailes:
So this has been deprecated since Jan, would the plan be to document the migration in the release notes?

Which part of the version number do you think we should change?

pankajagrawal16:
> So this has been deprecated since Jan, would the plan be to document the migration in the release notes?
> 
> Which part of the version number do you think we should change?

Yes, the plan is to document it as part of the release note and change log.  I am still considering a minor version increase since this should cause compile time issue for users

pankajagrawal16:
@msailes Have a look and see if you agree :)

msailes (approved):
👍

PR #395, opened 2021-05-14
feat: logger-Remove custom keys interface

### LINKED ISSUE #394: Ability to remove custom key from logger via LoggingUtils
Hi all, 
I have a question (or maybe a suggestion) about the power-tools-logging and pardon me if this is not the right place for this question. 
I recently come across this Lambda Power logging and its feature of changing the logging level via environment variable is a useful one for us. 
When I try to use it, I have coded to clear the custom key-pair values with ThreadContext.remove() and ThreadContext.clearAll() to avoid incorrect log output. I saw the power-tools logging LogginUtils contains method to appendKey() and appendKeys(), which in term calls ThreadContext.put() and putAll(). However, there is no wrapper methods to remove the keys. 
Hence, I have to still directly import the ThreadContext class to perform the remove operation in the code. 
It would be nice if LoggingUtils also include remove keys method so that my lambda code can completely rely on power-tools-logging for custom keys logging.

_Originally posted by @seng-thebouqs in https://github.com/awslabs/aws-lambda-powertools-java/discussions/393_

pankajagrawal16:
**Issue #, if available:** #394 

closes #394 

## Description of changes:

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/docs/core/logging.md
+++ b/docs/core/logging.md
@@ -154,6 +154,40 @@ You can append your own keys to your existing logs via `appendKey`.
     }
     ```
 
+
+### Removing additional keys
+
+You can remove any additional key from entry using `LoggingUtils.removeKeys()`.
+
+=== "App.java"
+
+    ```java hl_lines="19 20"
+    /**
+     * Handler for requests to Lambda function.
+     */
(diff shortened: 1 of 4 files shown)


msailes (approved):
👍

PR #484, opened 2021-07-27
chore: Logging and SQS utility Optimisations

pankajagrawal16:
**Issue #, if available:**

## Description of changes:

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
+++ b/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
@@ -56,19 +56,19 @@ public final class LambdaHandlerProcessor {
                 && pjp.getArgs()[2] instanceof Context;
     }
 
-    public static Optional<Context> extractContext(final ProceedingJoinPoint pjp) {
+    public static Context extractContext(final ProceedingJoinPoint pjp) {
 
         if (isHandlerMethod(pjp)) {
             if (placedOnRequestHandler(pjp)) {
-                return of((Context) pjp.getArgs()[1]);
+                return (Context) pjp.getArgs()[1];
             }
 
             if (placedOnStreamHandler(pjp)) {
-                return of((Context) pjp.getArgs()[2]);
+                return (Context) pjp.getArgs()[2];
             }
         }
 
-        return empty();
+        return null;
     }
 
     public static String serviceName() {
--- a/powertools-logging/src/main/java/software/amazon/lambda/powertools/logging/LoggingUtils.java
+++ b/powertools-logging/src/main/java/software/amazon/lambda/powertools/logging/LoggingUtils.java
@@ -26,7 +26,7 @@ import static java.util.Arrays.asList;
  * {@see Logging}
  */
 public final class LoggingUtils {
-    private static ObjectMapper objectMapper = new ObjectMapper();
+    private static ObjectMapper objectMapper;
 
     private LoggingUtils() {
     }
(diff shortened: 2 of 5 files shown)


powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
```diff
@@ -56,19 +56,19 @@ public static boolean placedOnStreamHandler(final ProceedingJoinPoint pjp) {
                 && pjp.getArgs()[2] instanceof Context;
     }
 
-    public static Optional<Context> extractContext(final ProceedingJoinPoint pjp) {
+    public static Context extractContext(final ProceedingJoinPoint pjp) {
```
pankajagrawal16, line 59: Avoiding too many usage of Optional

powertools-logging/src/main/java/software/amazon/lambda/powertools/logging/LoggingUtils.java
```diff
@@ -93,6 +93,10 @@ public static void defaultObjectMapper(ObjectMapper objectMapper) {
     }
 
     public static ObjectMapper objectMapper() {
+        if(null == objectMapper) {
```
pankajagrawal16, line 96: Lazy load object mapper. We dont want to instantiate mapper if Logger is not configured to log payload

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
```diff
@@ -231,6 +231,10 @@ public static void overrideSqsClient(SqsClient client) {
                                              final SqsMessageHandler<R> handler) {
         final List<R> handlerReturn = new ArrayList<>();
 
+        if(client == null) {
```
pankajagrawal16, line 234: Lazy load SQS client, we dont want to initialise it beforehand since users might be overrriding it as well.

msailes: We could improve the creation of the SqsClient?            

 .region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))

for example?

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
msailes: is this a problem?

pankajagrawal16: Yeah not generally .. but i am trying to avoid any extra computations .. Intresting blog which i referred before when working on it https://struberg.wordpress.com/2017/01/28/optional-vs-if-null/..

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
pankajagrawal16: Yeah its a possibilty, rite now I just converted to lazy load coz we have ability to specify overrriden client, want me to update ?

msailes: I'm wondering if it might cause problems for local running.

pankajagrawal16: I believe it can, but the reason why I avoided optimising default client was that user always have an ability to override it to most optimized way but then again it might not be obvious for new users to do so

msailes: yeah, ok.

msailes (approved):
👍

Issue #558, opened 2021-09-27
Java-based cfn-response module


bdkosher:
**Is your feature request related to a problem? Please describe.**
There's a [cfn-response module](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html) to support Lambda-backed AWS CloudFormation custom resources, but it's only available for Node and Python runtimes.

**Describe the solution you'd like**
A cfn-module equivalent for Java.

**Describe alternatives you've considered**
Lambda authors must create their own Java equivalent or use a supported runtime.

msailes:
I think this would be helpful to customers, and it would save them a task.

Even better if we could then update the docs.

heitorlessa:
cc @richardhboyd in case that's been discussed before somewhere else

PR #560, opened 2021-09-28
Create Java cfn-response equivalent (#558)

bdkosher:
Introduce Java implementation of the [cfn-response module](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html) based on NodeJS design.

**Issue #, if available:**

Addresses #558

## Description of changes:

* Add new powertools-cloudformation Maven module to house the sole Java class, `CloudFormationResponse`.
* Include one new parent POM dependency on `software.amazon.awssdk:http-client-spi:${aws.sdk.version}`.
* Add Unit Test class with 100% code coverage.

**Checklist**

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

None. New code.

**RFC issue #**:


bdkosher:
Is the [SpotBugs / codecheck failure](https://github.com/awslabs/aws-lambda-powertools-java/pull/560/checks?check_run_id=3733498206) a false positive? 

The logs do not show any violations for the impacted module:
`/home/runner/work/aws-lambda-powertools-java/aws-lambda-powertools-java/powertools-cloudformation/target/spotbugsXml.xml has 0 violations`

msailes:
A couple of questions from me.
- Do you think it would be useful to include a builder for `ResponseBody` this would give customers a similar feel to using the v2 SDK and other AWS Java libraries?
- Can you add documentation for this module so it appears on the website - https://awslabs.github.io/aws-lambda-powertools-java/

bdkosher:
1. `ResponseBody` is currently more of an internal object for producing the desired JSON payload. We _could_ directly expose that object for maximal control and have that be passed to a singular "send" method instead of having various polymorphic "send"s with different parameters. Alternatively, we could make the overall class follow the builder pattern, culminating in a "build" that would actually perform the "send". I do like the idea of having a builder here one way or another. I'm not particularly fond that I cannot specify a `noSend` arg without having to explicitly specify a `physicalResourceId` arg as well.
2. Certainly! I totally didn't notice the `docs` folder earlier.

Another design issue for your consideration: Right now there's a single `ObjectMapper` that arguably does double-duty: format the overall JSON in the HTTP request _and_ format the `Object responseData` that gets included in that payload. 

For my own application, e.g. I would ideally like to pass a `software.amazon.awssdk.services.chime.model.CreateAppInstanceResponse` as the "responseData", but this class does not serialize nicely. I could pass in my own `ObjectMapper` that would support serializing this type, but then that might interfere with overall JSON serialization (which is why I've currently made the constructor that accepts the ObjectMapper arg protected). Some ideas:
* Instead of a generic `Object` arg, use `Map<String, Object>` or possibly `Map<String, String>` to better mimic the "list of name value pairs" described in the JS/Python versions.
* Allow a custom ObjectMapper to be provided on init, but keep it separate from the generic ObjectMapper used to produce the overall JSON (might be a bit tricky to implement)
* Or just leave things the way they are.

pankajagrawal16:
> Is the [SpotBugs / codecheck failure](https://github.com/awslabs/aws-lambda-powertools-java/pull/560/checks?check_run_id=3733498206) a false positive?
> 
> The logs do not show any violations for the impacted module: `/home/runner/work/aws-lambda-powertools-java/aws-lambda-powertools-java/powertools-cloudformation/target/spotbugsXml.xml has 0 violations`

Yeah I have been trying to figure this one out. It looks like it fails on PRs where author donot have access to repo. I need to find some time to troubleshoot this further.

powertools-cloudformation/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-lambda-java-events</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
```
pankajagrawal16, line 61: Is this needed ?

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
```diff
@@ -0,0 +1,249 @@
...
+        private final boolean noEcho;
+        private final Object data;
+
+        ResponseBody(CloudFormationCustomResourceEvent event,
+                     Context context,
+                     ResponseStatus responseStatus,
+                     Object responseData,
+                     String physicalResourceId,
+                     boolean noEcho) {
+            requireNonNull(event, "CloudFormationCustomResourceEvent");
```
pankajagrawal16, line 57: Can we use the method from Objects class itself ? which has requireNonNull

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
```diff
@@ -0,0 +1,249 @@
...
+
+    private final SdkHttpClient client;
+    private final ObjectMapper mapper;
+
+    /**
+     * Creates a new CloudFormationResponse that uses the provided HTTP client and default JSON serialization format.
+     *
+     * @param client HTTP client to use for sending requests; cannot be null
+     */
+    public CloudFormationResponse(SdkHttpClient client) {
```
pankajagrawal16, line 120: Will be good to provide an implementation where user dont need to care about passing http client as well and module can take care of setting it up.

From Custom resource perspective, I believe it will provide better developer experience.

pankajagrawal16:
@bdkosher Firstly, Thanks for opening the issue and PR.

I am thinking out loud here but will it be better developer experience if we provide abstract implementation like this:

```
public abstract class CloudformationCustomResource {
    
    public Object handleRequest(CloudFormationCustomResourceEvent input, Context context) {
        Response response = null;

        switch (input.getRequestType()) {
            case "CREATE":
                response = create(input, context);
                buildAndSendCfnResponse(response, input, context);
                break;
            case "UPDATE":
                response = update(input, context);
                buildAndSendCfnResponse(response, input, context);
                break;
            // and so on
        }

        return response;
    }

    protected void buildAndSendCfnResponse(Response response, CloudFormationCustomResourceEvent input, Context context) {
        // Map Response to Response Body
        // Send response
    }

    public abstract Response create(CloudFormationCustomResourceEvent input, Context context);

    public abstract Response update(CloudFormationCustomResourceEvent input, Context context);

    public abstract Response delete(CloudFormationCustomResourceEvent input, Context context);
}
```

Then from builder perspective, we can control alot of things, for example, `Response` object can have all the builder goodness and be very specific to what we expect users to provide us. It can be a good place to capture mapper as which they would want us  to use to serialize response data. 

Also then builders wont need to know too much details about how to handle custom resource events etc. From their perspective, they just need to do their stuff for create, update and delete and provide us a response object.

What do you think? I believe this is closer to the implementation we have for python one.

Builder experience then looks something like this:

```
public class CustomResourceHandler extends CloudformationCustomResource {


    @Override
    public Response create(CloudFormationCustomResourceEvent input, Context context) {
        return null;
    }

    @Override
    public Response update(CloudFormationCustomResourceEvent input, Context context) {
        return null;
    }

    @Override
    public Response delete(CloudFormationCustomResourceEvent input, Context context) {
        return null;
    }
}
```

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
bdkosher: Good call.

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
bdkosher: True. The JS and Python versions embed the client. We'd just need to 
1. Pick a particular client impl (Apache?) and 
2. Include it as a dependency to the project.

powertools-cloudformation/pom.xml
bdkosher: I'm using a Jackson ObjectMapper to produce the JSON request body.

pankajagrawal16: I meant aspectj sorry

bdkosher: The module code itself does not use AspectJ, but when running `mvn package` from the parent POM, the `aspectj-maven-plugin` runs for every module and, if a module does not include that dependency, it'll fail the build:
```
[INFO] --- aspectj-maven-plugin:1.14.0:compile (default) @ powertools-cloudformation ---
[INFO] Showing AJC message detail for messages of types: [error, warning, fail]
[WARNING] couldn't find aspectjrt.jar on classpath. checked: ....
[ERROR] classpath error: unable to find org.aspectj.lang.JoinPoint (check that aspectjrt.jar is in your classpath)
```
I tried adding org.aspectj:aspectjrt as a plugin dependency but that didn't resolve the error.

bdkosher:
Seems like a good idea; as far as I know, CloudFormationResponse only gets used in the context of a handler that does some sort of one-time stack setup action based on the event type. So if I understand you correctly, we'd add two more classes:

1. An abstract base handler class that would allow subclasses to just "do their stuff for create, update and delete and provide us a response object"
   1. Should we have the create/update/delete methods call send() with ResponseStatus = SUCCESS, Response = null by default? Subclasses would only need to override the methods they care about.
   2. What's the proper way to handle the situation when communication to the custom resource fails? Invoke some other method that subclasses could override if they wanted to?
   3. Should this class itself implement RequestHandler?
2. A custom `Response` type to model the response data included in the send payload, plus a nested Builder for that type.

I'm debating between having `Response` hold name/value pairs vs. essentially be a wrapper for some `java.lang.Object` with option to customize its ObjectMapper. I know for my application's case, it's more convenient to manually convert my response object to a Map than configuring a custom ObjectMapper.

```java
void sendResponse(event, context, status, CreateAppInstanceResponse chimeResponse) {
    // easier to do this than configure a mapper to serialize artifact object type
    Map<String, String> data = Map.of("AppInstanceArn", chimeResponse.appInstanceArn());
    new CloudFormationResponse(ApacheHttpClient.create()).send(event, context, status, data);
}
```

pankajagrawal16:
> Should we have the create/update/delete methods call send() with ResponseStatus = SUCCESS, Response = null by default? Subclasses would only need to override the methods they care about.

Idea was to basically let users return a custom response object which we provide with builders etc. Users can decide based on the processing what will be the status etc. But basically we accept most of the parameters (Some optional probably) which users would want to send in response. This will include a custom object as well as part of response. The response object can then as well provide a method to accept custom object mapper which they would probably want to use when serializing response object. Builder dont need to then worry about sending the repsonse itself. That will be taken care by the handler which we implement as part of `CloudformationCustomResource ` . Makes sense ?

pankajagrawal16:
> What's the proper way to handle the situation when communication to the custom resource fails? Invoke some other method that subclasses could override if they wanted to?

It might be an idea to give an optional hook which is called in case send response invocation fails. Although I wonder what would user do in that case ? But still it might be a good idea to provide that flexibility .

pankajagrawal16:
> Should this class itself implement RequestHandler?

That was my thought yes. So Builder just need to provide implementation of three methods and return the response object.

pankajagrawal16:
> I'm debating between having Response hold name/value pairs vs. essentially be a wrapper for some java.lang.Object with option to customize its ObjectMapper. I know for my application's case, it's more convenient to manually convert my response object to a Map than configuring a custom ObjectMapper.

I believe it can accept any object, which could be a map as well if they want it to be. We can document that by default it will be serialized using default instance of object mapper, but if they want to use specific object mapper configuration, they can provide one as part of the custom Response object.

pankajagrawal16:
> > Is the [SpotBugs / codecheck failure](https://github.com/awslabs/aws-lambda-powertools-java/pull/560/checks?check_run_id=3733498206) a false positive?
> > The logs do not show any violations for the impacted module: `/home/runner/work/aws-lambda-powertools-java/aws-lambda-powertools-java/powertools-cloudformation/target/spotbugsXml.xml has 0 violations`
> 
> Yeah I have been trying to figure this one out. It looks like it fails on PRs where author donot have access to repo. I need to find some time to troubleshoot this further.

Try rebasing with master and that should now fix spotbugs check issue for now.

bdkosher:
Pushed some changes for your consideration. 

I added the abstract resource handler and Response classes (with Builder) as we discussed, as well as a `ResponseException` (I wasn't feeling particularly creative with the name so I'm open to alternatives). The `CloudFormationResponse` is still there. It had been named to mimic the "cfn-response" module, but now with these news classes I'm thinking we should rename it to `ResponseSender` or something along those lines.

`ResponseException` is used to differentiate between an Exception occurring while constructing a response (ResponseException) vs. sending a response (IOException). Constructing the response can fail in two ways:

1. There's a RuntimeException thrown by the create/update/delete method of the handler
2. There's some RuntimeException/JacksonException thrown while attempting to merge the ResponseBody wrapper object with the Response object that ends up as the "Data" property. 

I'm not fond of the the nested try/catches being used to handle these two failure scenarios--maybe there's a way to refactor `CloudFormationResponse` to just deal with "response sending" concerns and put the "response construction concerns" elsewhere that could be handled by a single try/catch.

If there's an exception while sending the response, I'm currently just logging it and invoking the empty-by-default `onSendFailure` method on the handler--there's not much you can really do at that point, as far as I know.

A final note, I'm still requiring the `SdkHttpClient` to be injected. I figured we could review these changes first and later figure out which HTTP client impl we wanted to use by default.

pankajagrawal16:
> Pushed another commit taking most of your feedback into consideration.
> 
> * I left `CloudFormationResource` public for the time being. Since other language modules effectively provide that object alone, I was hesitant to make it inaccessible for Java classes.
> * The abstract Handler class still requires an SdkHttpClient arg at present. Still unsure about adding a default impl and dependency.
> * No documentation updates included as of yet. I plan to add them to the PR once we finalize the APIs.

I think everything looks much better. This is seems to shape up real nice now. 

I still believe it's better to close `CloudFormationResource` down. See it's easier to open it up based on community feedback and hearing use cases around it later than starting with an open one. Then it's impossible to close it ever.

For Default client, I think especially for the use case of Custom resource, Cold start is a big problem, so if I have to choose taking a extra dep over better builder experience and ease of use, I would say latter wins :)

pankajagrawal16:
> True. OK with "software.amazon.awssdk:url-connection-client" as the default client? Looks to be the smallest client size-wise.

Sounds good

PR #577, opened 2021-10-11
Some code suggestion from CodeGuru

XinyuLiu5566:
- The code is using ConcurrentHashMap, but the usage of containsKey() and get() may not be thread-safe. In between the check and the get() another thread can remove the key and the get() will return null. Fix: 
Consider calling get(), checking instead of your current check if the returned object is null, and then using that object only, without calling get() again.

- This code uses '%s' to format int: 'size' expression. Use %d, not %s, for integers. This ensures locale-sensitive formatting.
- Replacing put() with putIfAbsent() to help prevent accidental overwriting. putIfAbsent() puts the value only if the ConcurrentHashMap does not contain the key and therefore avoids overwriting the value written there by the other thread's putIfAbsent().

### DIFF
--- a/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/DataStore.java
+++ b/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/DataStore.java
@@ -47,7 +47,7 @@ public class DataStore {
 
     public Object get(String key) {
         ValueNode node = store.get(key);
-        return node ? node.value : null;
+        return node != null ? node.value : null;
     }
 
     public boolean hasExpired(String key, Instant now) {


powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/cache/DataStore.java
```diff
@@ -46,7 +46,8 @@ public void remove(String Key){
     }
 
     public Object get(String key) {
-        return store.containsKey(key)?store.get(key).value:null;
+        ValueNode node = store.get(key);
+        return node ? node.value : null;
```
michaelbrewer, line 50: @XinyuLiu5566  - this would fix the runtime error

```suggestion
        return node != null ? node.value : null;
```

kozub:
Thanks @jeromevdl for merging #984. Could you please close this PR, as it contains the same changes and my PR was created just to fix some issues with this one.

jeromevdl:
Fixed with #984

PR #803, opened 2022-03-23
chore: add owasp vulnerability check

jeromevdl:
After the log4j event in dec/jan, good to have a check

**Issue #, if available:**

## Description of changes:

Added the owasp dependency-check plugin that scan dependencies for known CVEs. Fails the build if a CVE is found with level > 8

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


pankajagrawal16:
@jeromevdl Curious how is it different from dependabot dependency upgrades which gets merged automatically too ?

jeromevdl:
In several personal repo, dependabot didn't detect the log4j version in December... So I'm a kind of suspicious about it and prefer to have another belt.

pankajagrawal16 (approved):

jeromevdl:
> @jeromevdl I wonder why there are no build checks that ran with this? 🤔

Yes, right. This is weird, as I only add a plug-in in the pom...

kozub:
Hey @jeromevdl and @pankajagrawal16, I had a look on this change and there are couple of topics that I would like to discuss: 

1. This plugin is attached to `verify` mvn's phase, so it will by only invoked by SpotBug and Publish builds ('Build' builds uses `package phase, which is earlier). 

2. Do we have dedicated machine for running our builds? Plugin documentations says `It is important to understand that the first time this task is executed it may take 20 minutes or more as it downloads and processes the data from the National Vulnerability Database (NVD)` - I have try to run it on my machine and it failed after couple of minutes due to issues with downloading the data. Rerun helped ;-) 

3.  Do you have any opinion about CVSS failing level? Right now it is 8 and it does not impact our build. When I changed it to 7, it started complaining about jackson-databind (CVE-2020-36518(7.5), CVE-2022-42003(7.5), CVE-2022-42004(7.5)). Level 7 is also considered as high risk.

jeromevdl:
@kozub, you are right, this is the verify phase. Maybe we should do that together wit the spotbug check (`verify` instead of `install`). 
For the rest, let's just try !

jeromevdl:
looks like it runs with spotbugs... don't know what happened.

But it is also not super reliable (error 504): https://github.com/jeremylong/DependencyCheck/issues/5702

scottgerring:
Closing due to overlap with existing tooling (dependabot, sonar, etc.)

PR #826, opened 2022-04-08
feat: add AppConfig as a parameter provider

jeromevdl:
**Issue #, if available:**

## Description of changes:

Add AppConfig in the parameter module.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/docs/utilities/parameters.md
+++ b/docs/utilities/parameters.md
@@ -180,7 +180,7 @@ or the client SDK.
 
 === "AppConfigProvider"
 
-    ```java hl_lines="5 8-9"
+    ```java hl_lines="5 9-10"
     import software.amazon.lambda.powertools.parameters.AppConfigProvider;
     import software.amazon.lambda.powertools.parameters.ParamManager;
 
@@ -236,7 +236,7 @@ ParametersFunction:
 
 === "AppConfigProvider"
 
-    ```java hl_lines="5 8-9"
+    ```java hl_lines="5 9-10"
     import software.amazon.lambda.powertools.parameters.AppConfigProvider;
     import software.amazon.lambda.powertools.parameters.ParamManager;
 
--- a/docs/utilities/parameters.md
+++ b/docs/utilities/parameters.md
@@ -4,9 +4,13 @@ description: Utility
 ---
 
 
-The parameters utility provides a way to retrieve parameter values from
-[AWS Systems Manager Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html) or
-[AWS Secrets Manager](https://aws.amazon.com/secrets-manager/). It also provides a base class to create your parameter provider implementation.
+The parameters utility provides a way to retrieve parameter values from:
+
+ - [AWS Systems Manager Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html)
+ - [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/)
+ - [AWS AppConfig](https://aws.amazon.com/systems-manager/features/appconfig/)
+
+It also provides a base class to create your parameter provider implementation.
 
 **Key features**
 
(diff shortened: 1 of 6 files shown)


docs/utilities/parameters.md
```diff
@@ -166,6 +171,86 @@ in order to get data from other regions or use specific credentials.
...
+ParametersFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: Function
+      Handler: org.demo.parameters.ParametersFunction::handleRequest
+      Layers:
+        - arn:aws:lambda:us-east-1:027255383542:layer:AWS-AppConfig-Extension:68
+      Environment:
+        Variables:
+          POWERTOOLS_APPCONFIG_EXTENSION: 'true'
```
pankajagrawal16, line 234: From user experience perspective, is there a way we can build this intelligence within the AppConfig Provider? Like figuring it out from function config etc ?

heitorlessa:
I read an env bar for Extension. Is there a RFC to learn more about this
behaviour?

For Python, we offer normal integration with AppConfig via the Parameters
utility. I couldn’t find a related issue or RFC in the PR hence the ask.

Thanks!

jeromevdl:
> I read an env bar for Extension. Is there a RFC to learn more about this behaviour? For Python, we offer normal integration with AppConfig via the Parameters utility. I couldn’t find a related issue or RFC in the PR hence the ask. Thanks!

Not sure to understand your point @heitorlessa . The aim of this PR is actually to add AppConfig to the Parameters module. There's no RFC, or whatever, just wanted to provide this feature which is available in python. I'm currently looking for another way to detect if the extension is available rather than using this env var. If you know how, let me know...

docs/utilities/parameters.md
```diff
@@ -166,6 +171,86 @@ in order to get data from other regions or use specific credentials.
...
+        // Get an instance of the AppConfig Provider
+        AppConfigProvider appConfigProvider = ParamManager.getAppConfigProvider(client);
+    
+        // Retrieve a single secret
+        String value = appConfigProvider.get("/app/prod/config");
+    
+    }
+    ```
+
+### Using the AppConfig Extension for Lambda
```
heitorlessa, line 216: This is the part I was referring to @jeromevdl - we don't support this in Python.

That's why I was asking for a RFC to understand the behaviour so it could be assessed for other languages later.

Does that make sense?

heitorlessa:
> > I read an env bar for Extension. Is there a RFC to learn more about this behaviour? For Python, we offer normal integration with AppConfig via the Parameters utility. I couldn’t find a related issue or RFC in the PR hence the ask. Thanks!
> 
> 
> 
> Not sure to understand your point @heitorlessa . The aim of this PR is actually to add AppConfig to the Parameters module. There's no RFC, or whatever, just wanted to provide this feature which is available in python. I'm currently looking for another way to detect if the extension is available rather than using this env var. If you know how, let me know... 

Np, I've added a comment on the exact line that suggests Powertools supports AppConfig Lambda Extension - we don't support that in Python today, hence the ask for a RFC.

For your question, AppConfig Lambda Extension runs locally on localhost:2772 - if the socket is open the customer is running AppConfig extension.

Under "How it works"

https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-integration-lambda-extensions.html

jeromevdl:
@heitorlessa @pankajagrawal16 
==> #827

After writing this RFC, I must admit I'm not so sure it's a good idea to provide this option.

pankajagrawal16:
> @heitorlessa @pankajagrawal16 ==> #827
> 
> After writing this RFC, I must admit I'm not so sure it's a good idea to provide this option.

Looking at drawbacks you mentioned in RFC, I would agree as well. So in my eyes, it will add value to support extension bit if we can technically make DX experience like if users adds app config extension to the function and if they were using this module as well, the functionality should just seamlessly switch to get data from cached extension environment.

heitorlessa:
> @heitorlessa @pankajagrawal16 ==> #827
> 
> After writing this RFC, I must admit I'm not so sure it's a good idea to provide this option.

this is the best part of writing, it helps us put things in perspective and to perhaps see it from a different angle. I agree too it's best to drop the Lambda Extension part for now. 

Future wise, this gives you a better chance to hear customer demand, and work with customers that do use the Extension to confirm whether **(1)** this would be beneficial, and **(2)** what a good DX looks like (trade-offs included).

jeromevdl:
Closing (as per the above comments)

Issue #959, opened 2022-10-25
Parameter envelope is not taken in consideration in @Validation on requestHandler accepting SQSEvent


chihebdaoues:
My team is trying to use `aws-lambda-powertools' validation utility to validate request/response schemas on a couple of lambda functions we have and we noticed that the envelope parameter is not being taken in consideration when provided.

**What were you trying to accomplish?**
We have an SQS queue that subscribes to an SNS topic and a lambda function that subscribes to the SQS queue. Which I think is fairly common pattern. I was trying to annotate the request handler with validate to put some validation on the event that originated from the SNS topic schema and as such I needed to use the envelope parameter.

```
@Validation(inboundSchema = "classpath:/schemas/requestSchema.json", envelope = "records[*].powertools_json(body).powertools_json(Message)")
    public Void requestHandler(SQSEvent event, Context context) {
....
```

## Expected Behavior
I expected the Validation to use the envelope provided, but I think powertools is actually using the envelope for SQSEvents. I was expecting the default behavior to be overridden by the provided envelope.

## Current Behavior
Looking at the code in [ValidationAspect.java#L108](https://github.com/awslabs/aws-lambda-powertools-java/blob/bcf9ababdfe64fda62ff023ee36a18825f976165/powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/internal/ValidationAspect.java#L108), I see that the provided envelope will be used as the last resort. And since my event is an SQSEvent, it evaluates to true on [ValidationAspect.java#L71](https://github.com/awslabs/aws-lambda-powertools-java/blob/bcf9ababdfe64fda62ff023ee36a18825f976165/powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/internal/ValidationAspect.java#L71) and as such my envelope is not being used.

The current behavior is enforcing a one solution fits all for validating all the [builtin events](https://awslabs.github.io/aws-lambda-powertools-java/utilities/validation/#built-in-events-and-responses).

## Possible Solution
Test if the envelope is not empty at the start of the `if...else` block and if so use it. This would allow the user to override the default behavior which enables clients to customize the validation as they want.

## Steps to Reproduce (for bugs)
1. Create an SNS topic
2. Subscribe to the topic using an SQS queue
3. Create a lambda that subscribes to the SQS queue
4. Try to validate the schema of message that originated from the SNS topic

## Environment

* **Powertools version used**: 1.12.2
[...]

PR #960, opened 2022-10-25
fix: envelope is not taken into account with built-in types

jeromevdl:
**Issue #, if available:** #959

## Description of changes:
When an envelope is set, it takes precedence over the built-in types. If not set, built-in types validation apply. If the type is not handled, a warn is logged.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

People using an envelope with a built-in type (useless) may now enter in the envelope validation rather than the built-in one, causing validation error.

**RFC issue #**:

### DIFF
--- a/powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/ValidationUtils.java
+++ b/powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/ValidationUtils.java
@@ -13,6 +13,7 @@
  */
 package software.amazon.lambda.powertools.validation;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collections;
@@ -21,9 +22,12 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
+import com.amazonaws.services.lambda.runtime.serialization.PojoSerializer;
+import com.amazonaws.services.lambda.runtime.serialization.events.LambdaEventSerializers;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeType;
+import com.fasterxml.jackson.databind.node.NullNode;
 import com.networknt.schema.JsonSchema;
 import com.networknt.schema.ValidationMessage;
 import io.burt.jmespath.Expression;
@@ -65,9 +69,15 @@ public class ValidationUtils {
         }
         JsonNode subNode;
         try {
-            JsonNode jsonNode = ValidationConfig.get().getObjectMapper().valueToTree(obj);
+            PojoSerializer pojoSerializer = LambdaEventSerializers.serializerFor(obj.getClass(), ClassLoader.getSystemClassLoader());
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            pojoSerializer.toJson(obj, out);
+            JsonNode jsonNode = ValidationConfig.get().getObjectMapper().readTree(out.toString("UTF-8"));
             Expression<JsonNode> expression = ValidationConfig.get().getJmesPath().compile(envelope);
             subNode = expression.search(jsonNode);
+            if (subNode == null || subNode instanceof NullNode) {
+                throw new ValidationException("Not found");
+            }
         } catch (Exception e) {
             throw new ValidationException("Cannot find envelope <"+envelope+"> in the object <"+obj+">", e);
         }
--- a/powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/ValidationUtils.java
+++ b/powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/ValidationUtils.java
@@ -76,7 +76,7 @@ public class ValidationUtils {
             Expression<JsonNode> expression = ValidationConfig.get().getJmesPath().compile(envelope);
             subNode = expression.search(jsonNode);
             if (subNode == null || subNode instanceof NullNode) {
-                throw new ValidationException("Not found");
+                throw new ValidationException("Envelope not found in the object");
             }
         } catch (Exception e) {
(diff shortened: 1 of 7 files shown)


powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/internal/ValidationAspect.java
```diff
@@ -36,6 +38,9 @@
  */
 @Aspect
 public class ValidationAspect {
+    private static final Logger LOG = LoggerFactory.getLogger(ValidationAspect.class);
+
+
```
msailes, line 43: extra new line

powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/internal/ValidationAspect.java
```diff
@@ -59,7 +64,9 @@ && placedOnRequestHandler(pjp)) {
                 JsonSchema inboundJsonSchema = getJsonSchema(validation.inboundSchema(), true);
 
                 Object obj = pjp.getArgs()[0];
-                if (obj instanceof APIGatewayProxyRequestEvent) {
+                if (validation.envelope() != null && !validation.envelope().isEmpty()) {
```
msailes, line 67: Could there be cases where customer functionality changes because of this reordering?

powertools-validation/src/test/java/software/amazon/lambda/powertools/validation/handlers/SQSWithCustomEnvelopeHandler.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.SQSEvent;
+import software.amazon.lambda.powertools.validation.Validation;
+
+public class SQSWithCustomEnvelopeHandler implements RequestHandler<SQSEvent, String> {
+
+    @Override
+    @Validation(inboundSchema = "classpath:/schema_v7.json", envelope = "records[*].powertools_json(body).powertools_json(Message)")
```
chihebdaoues, line 24: One last thing, notices how records in the envelope JMSPath starts with a lower-case `r` but that in the schema it starts with an upper-case `R`. Also, on your website documentation for builtin events, you're specifying the path with a capital `R`. When I was trying to build my custom envelope starting from the builtin events didn't really help. Also I am not sure but it's worth checking if JMSPaths are case-sensitive.

jeromevdl: You are actually right, the issue is that using the [aws-lambda-java-events](https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-events) library, [SQSEvent](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/SQSEvent.java) has a "records" (r) field while the [true event](https://docs.aws.amazon.com/fr_fr/lambda/latest/dg/with-sqs.html) has a "Records" (R). 

The validation library uses the event library, which is actually wrong when talking about JMESPath which looks at the real JSON... The test should have `envelope = "Records[*].powertools_json(body).powertools_json(Message)"` with big R. I can change this to reflect the reality. We may end up with breaking changes. @msailes wdyt ?

chihebdaoues: Hi Jerome,
I don't think that updating the test to have a big `R` would work, not sure maybe worth to check. I was more thinking that updating the documentation to have a small `r` would be the simpler option here.

 In any case, I don't think this is related to this PR, so I think this PR does need to be updated.

jeromevdl: Actually I was thinking about correcting the code to match the truth (R) and make the test with R pass. Small r should not work as the real event has big R...

msailes: I agree with you @jeromevdl

chihebdaoues (dismissed):
Thank you @jeromevdl, I can confirm this now works as expected. I tested it, creating events with the following code 
```
SQSEvent.SQSMessage sqsMessage = new SQSEvent.SQSMessage();
sqsMessage.setBody(messageBody);
```
And it works with a big R

powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/ValidationUtils.java
```diff
@@ -65,9 +69,15 @@ public static void validate(Object obj, JsonSchema jsonSchema, String envelope)
...
         try {
-            JsonNode jsonNode = ValidationConfig.get().getObjectMapper().valueToTree(obj);
+            PojoSerializer pojoSerializer = LambdaEventSerializers.serializerFor(obj.getClass(), ClassLoader.getSystemClassLoader());
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            pojoSerializer.toJson(obj, out);
+            JsonNode jsonNode = ValidationConfig.get().getObjectMapper().readTree(out.toString("UTF-8"));
             Expression<JsonNode> expression = ValidationConfig.get().getJmesPath().compile(envelope);
             subNode = expression.search(jsonNode);
+            if (subNode == null || subNode instanceof NullNode) {
+                throw new ValidationException("Not found");
```
msailes, line 79: Can this error message be expanded to be clearer to a dev that hits it?

msailes (approved):
LGTM!

powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/ValidationUtils.java
jeromevdl: Just changed the message. But the exception is catch and the main message is just below, with more details.

PR #971, opened 2022-11-04
fix!: bump metrics version to 4.0.3

jeromevdl:
**Issue #, if available:** #944

## Description of changes:

Update `aws-embedded-metrics` version to v4.0.3 (latest) and adjust the code to handle this bump (from v1.0.6)

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**BREAKING CHANGE**

This new version of `aws-embedded-metrics` introduce Checked Exception for multiple methods (`metricsLogger.putMetric()`, `DimensionSet.of()`, ...). Lambda Powertools internally catch them and throws Unchecked Exceptions, but when using the library itself in the Lambda handler, it is now required to handle this exception.

Example:

Before: 
```
metricsLogger.putMetric("CoolMetric", 1);
```

After:
```
try {
    metricsLogger.putMetric("CoolMetric", 1);
} catch (InvalidMetricException e) {
    log.error(e);
}
```

Questions: 
- should powertools-metrics also throw these checked exceptions?
- Or on the contrary, should powertools-metrics provide a wrapper for the MetricsLogger and hide these exceptions?

Developer Experience is quite bad with these checked exception, with a lot of try catch everywhere in the code...

**RFC issue #**: #944

### DIFF
--- a/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspect.java
+++ b/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspect.java
@@ -7,6 +7,10 @@ import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
 import org.aspectj.lang.annotation.Aspect;
 import org.aspectj.lang.annotation.Pointcut;
+import software.amazon.cloudwatchlogs.emf.exception.DimensionSetExceededException;
+import software.amazon.cloudwatchlogs.emf.exception.InvalidDimensionException;
+import software.amazon.cloudwatchlogs.emf.exception.InvalidMetricException;
+import software.amazon.cloudwatchlogs.emf.exception.InvalidNamespaceException;
 import software.amazon.cloudwatchlogs.emf.logger.MetricsLogger;
 import software.amazon.cloudwatchlogs.emf.model.DimensionSet;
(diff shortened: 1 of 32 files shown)


powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspect.java
```diff
@@ -83,12 +87,24 @@ private void coldStartSingleMetricIfApplicable(final String awsRequestId,
...
+            MetricsLogger metricsLogger = new MetricsLogger();
+            try {
                 metricsLogger.setNamespace(namespace(metrics));
+            } catch (InvalidNamespaceException e) {
+                throw new RuntimeException("A valid namespace is required, either pass it to the @Metrics annotation or set the environment variable POWERTOOLS_METRICS_NAMESPACE", e);
+            }
+            try {
                 metricsLogger.putMetric("ColdStart", 1, Unit.COUNT);
+            } catch (InvalidMetricException e) {
+                // should not occur
```
kozub, line 99: How about adding sole logs on debug level? Too many times in my live I've seen things that should not occur ;)

kozub (commented):
We had a couple of days ago a quick discussion about this change with @msailes and he was planing to discuss it also with metrics maintainers. Mark - any comments?

jeromevdl:
Switching destination branch to v2, as this cannot be merged into master (v1) because of breaking changes.

Issue #965, opened 2022-10-29
RFC: Switch to logback, SLF4J and logstash as logging framework


roamingthings:
## Key information

* RFC PR: 
* Related issue(s), if known:
* Area: Logger
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-java/#tenets): Yes

## Summary
[summary]: #summary

Switch to [logback](https://logback.qos.ch) [GitHub repository](https://github.com/qos-ch/logback), [SLF4J](http://www.slf4j.org) and [logstash](https://github.com/logfellow/logstash-logback-encoder) as logging framework used by the Logging module.

## Motivation
[motivation]: #motivation

Currently the Lambda Powertools are using log4j2 as logging framework. Log4j2 is a very popular and widely used logging library in the Java ecosystem. It offers a lot of features that are most likely not required by most Lambda functions. One of these features (JNDI support) has gained a lot of popularity through CVE-2021-44228.

Another popular logging framework that derived from log4j1 is [logback]((https://logback.qos.ch). When using logbook it's also required to use SLF4J as an abstraction layer. For example logback is the default logging framework used by Spring Boot. It has a smaller jar-artefact footprint than log4j2.

[logstash](https://github.com/logfellow/logstash-logback-encoder) is another library that adds JSON formatting and structured logging to logback.

This would allow applications to add structured logging in a clean and convenient way without adding and removing additional keys explicitly (see [proposal](#proposal) for an example). This is important because it reduces boilerplate code and the business logic gets more visibility.

## Proposal
[proposal]: #proposal

Replace the current usage of log4j2 with logback, SLF4J and also add logstash.

The interface of the Logging module would not have to change much. However, the configuration of the layout has to be changed to use logback and logstash. 

As stated in the [motivation section](#motivation) the usage of logstash would allow for clean implementation when using structured logging.

For example
```java
LoggingUtils.appendKey("foo", "bar");
log.info("Hello World!")
LoggingUtils.removeKey("foo");
```
would become
```java
log.info("Hello World! {}", kv("foo", "bar"));
```
Beside this common use-case logstash offers other convenient methods to generate these structured arguments.  [Structured Logging with Structured Arguments](https://www.innoq.com/en/blog/structured-logging/) for more details

##### Logback
[Dual license model](https://logback.qos.ch/license.html):
Eclipse Public License v1.0
GNU Lesser General Public License version 2.1

##### SLF4J
MIT license

##### Logstash
Apache 2

[...]

Issue #752, opened 2022-02-16
Consistent log keys across all runtimes


pankajagrawal16:
**Is your feature request related to a problem? Please describe.**

Python and Typescript use snake_case for all outputs.  Java uses a mix of snake_case and camel_case.

 Customers will use the same tool (CloudWatch Logs, Loggly, Datadog, etc.) to search data across logs, and if main keys are different they will find half of the information when using multiple runtimes.

At the sametime, if we break language consistency and the other application(not using powertools but standard logging libs) loose that tracking on single visibility. 

Considering both the above rationale, for now, we should standardize on the same key name (snake_case) for logging output - we, Golang will also using snake_case albeit in the early phase. As it’s a breaking change for Java, let’s strive for making it configurable but a v2 should standardize it on snake_case.

**Describe the solution you'd like**
Ability to configure if customers want consistent logging keys same as used in powertools logging for other runtimes.

**Describe alternatives you've considered**

None

**Additional context**

scottgerring:
#1435

PR #995, opened 2022-11-22
Feat!: remove log4j

jeromevdl:
**Issue #, if available:** #965

## Description of changes:
- Remove log4j as the de facto logging library
- Replace with SLF4J to give the ability to choose a logging framework
- Creation of 2 new modules (logging-log4j and logging-logback). These modules provide logging implementation. Others can be created (reload4j, JUL, ...), depending on the demand.
- logged fields now match python and typescript versions (ex: function_name, cold_start, ... instead of functionName, coldStart) to simplify searching and troubleshooting across different runtimes (see #752).

## Developer experience
The developer experience doesn't change that much. Instead of using the logging module, developers need to add the implementation module (logging-logback, logging-log4j, ...) which has a dependency on the logging module.

Ex:
```
        <dependency>
            <artifactId>powertools-logging-logback</artifactId>
            <version>1.12.3</version>
        </dependency>
```

Then they need to have the proper config file (log4j2.xml or logback.xml). 
- log4j2 is relying on the same mechanism as v1, using a `JsonTemplateLayout`. Two layout.json files are provided (standard and ECS). This can be extended. Developers can also choose to remove some fields generated by powertools or to rename them. The Resolver is more granular (each field is accessible).
```
<?xml version="1.0" encoding="UTF-8"?>
<Configuration>
    <Appenders>
        <Console name="JsonAppender" target="SYSTEM_OUT">
            <JsonTemplateLayout eventTemplateUri="classpath:LambdaJsonLayout.json" />
        </Console>
    </Appenders>
    <Loggers>
        <Root level="info">
            <AppenderRef ref="JsonAppender"/>
        </Root>
    </Loggers>
</Configuration>
```
- logback is relying on encoders. Two encoders are provided (`LambdaJsonEncoder` and `LambdaEcsEncoder`). We can probably go further in the customization here (for example having an 'excludeFields' attribute to remove some of the fields or an 'includeFields' to only select a few ones:
```
<configuration>
    <appender name="JsonAppender" class="ch.qos.logback.core.ConsoleAppender">
        <encoder class="software.amazon.lambda.powertools.logging.LambdaJsonEncoder" />
    </appender>

    <root level="INFO">
        <appender-ref ref="JsonAppender" />
    </root>
</configuration>
```

**Checklist**

* [] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-java/#tenets)
==> this version will not be backward compatible. Several changes are required from the developer, and it comes with several changes (field names)
==> but it comes with a more standard way of logging (slf4j) and does not force a specific framework. It is also a bit more open for extension.
* [ ] Update tests
==> not fully, wanted to share the DX first.
* [ ] Update docs
==> not yet, wanted to share the DX first.

## Breaking changes

- Need to import one of the logging implementations (in mvn/gradle), rather than the logging module.
- Logged fields are now using snake_case, to match Python and Typescript versions.
- This new version logs less information than before (to match Python and Typescript versions and also to reduce the cloudwatch logs size). Note that with the log4j implementation, it is possible to extend and add fields (using resolvers). Would be great to provide an equivalent for logback. 
- Developer needs to use Logger and LoggerFactory from SLF4J instead of log4j implementations.

**RFC issue #**: #965, #752


powertools-logging-logback/src/main/java/software/amazon/lambda/powertools/logging/LambdaEcsEncoder.java
```diff
@@ -0,0 +1,94 @@
...
+    private final ThrowableProxyConverter throwableProxyConverter = new ThrowableProxyConverter();
+    protected ThrowableHandlingConverter throwableConverter = null;
+
+    @Override
+    public byte[] headerBytes() {
+        return null;
+    }
+
+    @Override
+    public byte[] encode(ILoggingEvent event) {
```
stevehouel, line 41: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This method calls 27 other methods. By comparison, 98% of the methods in the CodeGuru reference dataset use fewer. This indicates the method is highly coupled with other methods. A method that is highly coupled with other methods is difficult to understand and its behavior might change unexpectedly when one of its referenced methods is updated. High coupling could also increase the integration test complexity, maintenance cost and technical debt.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code blocks on lines 56-65, lines 68-74 into separate methods.

jeromevdl:
@pankajagrawal16 I'd like to get your thought on this one if you have some time.

pankajagrawal16:
> @pankajagrawal16 I'd like to get your thought on this one if you have some time.

👍 I can give it a look sometime this week.

PR #1000, opened 2022-11-26
feat(metrics): introduce  MetricsUtils.withMetricsLogger utility

### LINKED ISSUE #999: metrics: Expose a MetricsUtils.withSingleMetric overload that doesn't require metric name/value/unit
**Is your feature request related to a problem? Please describe.**

`MetricsUtils.withSingleMetric` - [MetricsUtils.java](https://github.com/awslabs/aws-lambda-powertools-java/blob/master/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java) - provides a nice utility to hide the creation, configuration and publishing of metrics. The existing overloads however, assume that the user needs to know the name/value/unit of at least the first metric to add. This is not always true. I'd like to provide a `Consumer<MetricLogger>` that will add metrics based on some conditional logic.

The main use case my team has

1. We're emitting a number of metrics - ~10 metrics - that all have the same config (e.g. dimensions) that's meant to be different from MetricsUtils.metricsLogger() i.e. we need a new metrics logger instance
1. The names of those metrics are generated dynamically - think loops, appending different constants - to generate those names

So, ideally, we create a new instance of a metrics logger, and then rely on to emit those 10 different metrics.

**Describe the solution you'd like**

Introduce a new method, an overload of `MetricsUtils.withSingleMetric` or even a new name `MetricsUtils.withMetric` as follows

```
public static void withMetric(Consumer<MetricsLogger> logger) {
        MetricsLogger metricsLogger = logger();

        try {
            metricsLogger.setNamespace(defaultNameSpace());
            captureRequestAndTraceId(metricsLogger);
            logger.accept(metricsLogger);
        } finally {
            metricsLogger.flush();
        }
    }
```

which would allow the user to freely configure and add metrics via their provided `Consumer<MetricsLogger>`

**Describe alternatives you've considered**

**Additional context**

humanzz:
closes #999

**Issue #, if available:** #999

## Description of changes:

introduce new `MetricsUtils.withMetricsLogger` utility method to enable higher level of customization to the metrics logger and eliminate the need to know the metric/configuration when calling the method and leave that instead to the `Consumer<MetricsLogger>`

The main use case my team has
1. We're emitting a number of metrics - ~10 metrics - that all have the same config (e.g. dimensions) that's meant to be different from `MetricsUtils.metricsLogger()` i.e. we need a new metrics logger instance
2. The names of those metrics are generated dynamically - think loops, appending different constants - to generate those names

So, ideally, we create a new instance of a metrics logger, and then rely on to emit those 10 different metrics.

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
+++ b/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
@@ -121,6 +121,26 @@ public final class MetricsUtils {
         }
     }
 
+    /**
+     * Provide and immediately flush a {@link MetricsLogger}. It will use the default namespace
+     * specified either on {@link Metrics} annotation or via POWERTOOLS_METRICS_NAMESPACE env var.
+     * It by default captures function_request_id as property if used together with {@link Metrics} annotation. It will also
+     * capture xray_trace_id as property if tracing is enabled.
+     *
+     * @param logger the MetricsLogger
+     */
+    public static void withMetric(final Consumer<MetricsLogger> logger) {
+        MetricsLogger metricsLogger = logger();
+
+        try {
+            metricsLogger.setNamespace(defaultNameSpace());
+            captureRequestAndTraceId(metricsLogger);
+            logger.accept(metricsLogger);
+        } finally {
+            metricsLogger.flush();
+        }
+    }
+
     public static DimensionSet[] getDefaultDimensions() {
         return Arrays.copyOf(defaultDimensions, defaultDimensions.length);
     }
--- a/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
+++ b/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
@@ -81,16 +81,10 @@ public final class MetricsUtils {
                                         final double value,
                                         final Unit unit,
                                         final Consumer<MetricsLogger> logger) {
-        MetricsLogger metricsLogger = logger();
-
-        try {
-            metricsLogger.setNamespace(defaultNameSpace());
+        withMetric(metricsLogger -> {
             metricsLogger.putMetric(name, value, unit);
-            captureRequestAndTraceId(metricsLogger);
             logger.accept(metricsLogger);
-        } finally {
-            metricsLogger.flush();
-        }
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


msailes (approved):
LGTM

msailes (dismissed):
LGTM

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
```diff
@@ -81,16 +81,10 @@ public static void withSingleMetric(final String name,
                                         final double value,
                                         final Unit unit,
                                         final Consumer<MetricsLogger> logger) {
-        MetricsLogger metricsLogger = logger();
-
-        try {
-            metricsLogger.setNamespace(defaultNameSpace());
+        withMetric(metricsLogger -> {
             metricsLogger.putMetric(name, value, unit);
```
scottgerring, line 85: `putMetric` is no longer within the `try`/`finally` block. As far as I can see it doesn't throw any runtime exceptions - did you dive into this to make sure?

kozub:
It looks also good for me.

The only concern that I have is about naming. Right now we have two methods:  `withSingleMetric()`  and `withMetric()`  - without the context I would have no idea what is the difference.  Not sure if `.withCustomMetric()` or `.withConfigurableMetric()` sounds better. What do you thing? I don't have strong opinion what will be more natural for end user.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
humanzz: The way I thought about this, is that whatever was happening in the `try` block, remains happening in a `try` block. It's just that it moved to the new `withMetric` method.

humanzz:
@kozub 

> It looks also good for me.
> 
> The only concern that I have is about naming. Right now we have two methods: `withSingleMetric()` and `withMetric()` - without the context I would have no idea what is the difference. Not sure if `.withCustomMetric()` or `.withConfigurableMetric()` sounds better. What do you thing? I don't have strong opinion what will be more natural for end user.

The names are also throwing me off a bit tbh as well. I am torn between
- singular `withMetric` 
- plural `withMetrics`
- `withMetricsLogger` to really be super explicit, which I think is more in line with `MetricsUtils.metricsLogger()` and is actually more telling about what it provides.

kozub:
Hey @humanzz 

I think that `.withMetricsLogger()` sound more natural than previous version. Thanks for that change. 
Idea of having `.withMetric()` and `.withMetrics()` is also interesting, but unfortunately it introduces some breaking changes so we won't be able to release it so quickly.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsUtils.java
scottgerring: You're right - I misread the lambda. Cool!

scottgerring (dismissed):

humanzz:
@kozub @scottgerring apologies folks. Had to make an additional commit to fix a documentation typo

kozub (approved):

msailes (approved):
Approved

PR #1058, opened 2023-02-20
chore: update how a request handler method is identified

### LINKED ISSUE #1056: Powertools method of identifying request handler method doesn't work with Handler class hierarchies
My code base has several Lambda handler classes. I'm implementing an abstract base handler class with some common behaviours. I then implement a lambda specific handler extending the abstract base handler.

```java
...
import software.amazon.codeguruprofilerjavaagent.RequestHandlerWithProfiling;
...

@Log4j2
public abstract class RequestHandler<I, O> extends RequestHandlerWithProfiling<I, O> {

    static {
        // Override the default dimensions to no dimensions by setting the defaultDimensions to an empty list
        MetricsUtils.defaultDimensions();
    }

    @Override
    @Tracing(segmentName = "Handler")
    @Logging(clearState = true)
    @Metrics(captureColdStart = false) // Disable coldStart metric as it requires non-empty dimensions
    public O requestHandler(I i, Context context) {
        try {
            return handle(i, context);
        } catch (Exception e) {
            ....
        }


       public abstract O handle(I i, Context context);
}
```

**What were you trying to accomplish?**

In a concrete handler, extending the abstract `RequestHandler` above, I wanted to use the `@Validation` utility

```java
public class ConcreteHandler extends RequestHandler<SQSEvent, Void> {
     ...
    @Override
    @Validation(inboundSchema = ..., envelope = ...)
    public Void handle(SQSEvent event, Context context) {
        ...
        return null;
    }
    ...
}
```

## Expected Behavior

The `@Validation` to work.

## Current Behavior

`@Validation` did not result in validating the request. When debugging, I found the following
1. `ValidationAspect` leverage's `core`'s `LambdaHandlerProcessor` to identify handler methods https://github.com/awslabs/aws-lambda-powertools-java/blob/master/powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/internal/ValidationAspect.java#L58 and skips validation if the method is identified as a handler
2. `LambdaHandlerProcessor` makes some strong assumptions about handler method name https://github.com/awslabs/aws-lambda-powertools-java/blob/master/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java#L39

The end result, is that `@Validation` under my setup becomes a no-op.

## Possible Solution

I'm reporting the above with an example of `@Validation` annotation, but I believe most other Powertools utilities do share a similar logic in their `*Aspect.java` classes to leverage `LambdaHandlerProcessor` for identifying handler methods.

A possible solution is to maybe simply rely on method signature only, rather than forcing a name as already implemented in `LambdaHandlerProcessor.placedOnRequestHandler()` and `LambdaHandlerProcessor.placedOnStreamHandler()` and to drop the `LambdaHandlerProcessor.isHandlerMethod()` check completely.

## Steps to Reproduce (for bugs)

1.
4.
5.
6.

## Environment

* **Powertools version used**: 1.14.0
* **Packaging format (Layers, Maven/Gradle)**:
* **AWS Lambda function runtime:**
* **Debugging logs**

> [How to enable debug mode](https://awslabs.github.io/aws-lambda-powertools-java/#debug-mode)**

```text
# paste logs here
```

humanzz:
- stop relying on handler method name
- leverage placedOnRequestHandler and placedOnStreamHandler methods instead

closes #1056

**Issue #, if available:** 1056

## Description of changes:

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
+++ b/powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessor.java
@@ -37,9 +37,7 @@ public final class LambdaHandlerProcessor {
     }
 
     public static boolean isHandlerMethod(final ProceedingJoinPoint pjp) {
-        return "handleRequest".equals(pjp.getSignature().getName()) ||
-                // https://docs.aws.amazon.com/codeguru/latest/profiler-ug/lambda-custom.html
-                "requestHandler".equals(pjp.getSignature().getName());
+        return placedOnRequestHandler(pjp) || placedOnStreamHandler(pjp);
     }
 
     public static boolean placedOnRequestHandler(final ProceedingJoinPoint pjp) {
--- a/powertools-core/src/test/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessorTest.java
+++ b/powertools-core/src/test/java/software/amazon/lambda/powertools/core/internal/LambdaHandlerProcessorTest.java
@@ -1,8 +1,14 @@
 package software.amazon.lambda.powertools.core.internal;
 
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.RequestStreamHandler;
 import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.Signature;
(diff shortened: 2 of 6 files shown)


jeromevdl (approved):
Looks good to me. It means we can fake an handler method, but I don't think that's a big issue, if the Context is at its place.

kozub (approved):

PR #1066, opened 2023-02-21
docs(plugin): fix mdocs and git revision plugin integration

machafer:
**Issue #, if available:** [https://github.com/awslabs/aws-lambda-powertools-java/issues/855](url)

## Description of changes:

**Checklist**

 

* [X] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-java/#tenets)
* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -70,6 +70,7 @@ markdown_extensions:
 copyright: Copyright &copy; 2021 Amazon Web Services
 
 plugins:
+  - git-revision-date
   - search
   - macros
 
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
           python-version: "3.8"
       - name: Capture branch and tag
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -6,11 +6,16 @@ on:
       - master
     paths:
       - 'docs/**'
+      - 'mkdocs.yml'
+      - 'Makefile'
+
   push:
     branches:
       - master
     paths:
       - 'docs/**'
+      - 'mkdocs.yml'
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       max-parallel: 4
       matrix:
         # test against latest update of each major Java version, as well as specific updates of LTS versions:
-        java: [8, 8.0.192, 11.0.x, 11.0.3, 12, 13, 15, 16, 17 ]
+        java: [8, 11, 16, 17 ]
```
jeromevdl, line 45: I would keep the list of versions we had

mkdocs.yml
```diff
@@ -70,6 +70,7 @@ markdown_extensions:
 copyright: Copyright &copy; 2021 Amazon Web Services
 
 plugins:
+  - git-revision-date
```
jeromevdl, line 73: I don't see how it can work... we removed it because of an error (https://github.com/awslabs/aws-lambda-powertools-java/issues/855). There's nothing in the PR that makes me think it will be different now...

.github/workflows/build.yml
machafer: Yep, makes sense. I was doing some tests with the new actions/setup-javaV3.

mkdocs.yml
machafer: I'm still doing some experiments. Including testing the actions/setup-python@v4. However, the flow works on branch execution.

jeromevdl: I've added the build doc workflow, looks like it passes... I guess we can move on or are you still testing things?

machafer: Yes, it's ready. Thanks for your help.

jeromevdl: why did you remove what I've added ? I think that make sense to test the doc build when mkdocs changes...

jeromevdl: Also be careful with forced pushes, can be useful sometimes but looks like you do it each time.

jeromevdl (approved):

Issue #1081, opened 2023-03-09
powertools-cloudformation custom resource exception handling


scottgerring:
## Expected Behavior

I expect [`powertools-cloudformation`](https://awslabs.github.io/aws-lambda-powertools-java/utilities/custom_resources/) to use the `physicalResourceId` sent by CloudFormation during `UPDATE` events when **(A)** none are explicitly set, or **(B)** a `RuntimeException` was thrown.

## Current Behavior

<details>
    <summary><strong>Expand for a primer in CloudFormation Custom Resources</strong></summary>

> Checkout this page for a complete explanation: [CloudFormation Custom Resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html)

For the context of this issue, we will focus on the three main high level resource lifecycle events: **CREATE**, **UPDATE**, **DELETE**.

For each event, a unique identifier called **Physical Resource ID** is used to control whether a resource needs to be created, updated, replaced, or deleted. From here on, we'll refer to this as `physicalResourceId`.

In this context, the typical workflow for a custom resource runs as follows:

* [**CREATE** event](#cloudformation-events) is invoked when a resource is being provisioned for the first time.
  * **Expected response**. CloudFormation expects a `physicalResourceId` in a successful custom resource response, where its uniqueness is determined by the custom resource itself. From here on, CloudFormation will send the same `physicalResourceId` for any subsequent operation related to this resource.

* [**UPDATE** event](#cloudformation-events) is invoked when a resource had a property changed. It sends the `physicalResourceId` previously known for this resource.
  * **Expected response**. For a successful update of the same resource, CloudFormation expects the same `physicalResourceId`. However, if `physicalResourceId` is different, CloudFormation recognizes it needs to dispatch a `DELETE` event to clean up the previous `physicalResourceId` it had stored - effectively a resource replacement.

* [**DELETE** event](#cloudformation-event) is invoked when a resource was either removed from the template or its stack is being deleted. It sends the `physicalResourceId` previously known for this resource.
  * **Expected Response**. CloudFormation expects a non-error response to be sure the resource was deleted successfully.

</details>

## Problem

For successful operations (no-exception), [`powertools-cloudformation`](https://awslabs.github.io/aws-lambda-powertools-java/utilities/custom_resources/) implicitly defines the `physicalResourceId` value with the CloudWatch Log Stream name.

[...]

PR #1082, opened 2023-03-09
fix(cloudformation-module): Use physicalResourceId when not provided by custom resource

msailes:
**Issue #, if available:**
https://github.com/awslabs/aws-lambda-powertools-java/issues/1081

## Description of changes:

Use physicalResourceId when not provided by custom resource.

**Checklist**

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

No breaking changes

Response.success() and Response.failed are deprecated

### DIFF
--- a/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/AbstractCustomResourceHandler.java
+++ b/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/AbstractCustomResourceHandler.java
@@ -66,8 +66,10 @@ public abstract class AbstractCustomResourceHandler
             LOG.error("Unable to generate response. Sending empty failure to {}", responseUrl, rse);
             try {
                 // If the customers code throws an exception, Powertools should respond in a way that doesn't
-                // change the CloudFormation resources. A failure is sent with the existing PhysicalResourceId
+                // change the CloudFormation resources.
+                // In the case of a Update or Delete, a failure is sent with the existing PhysicalResourceId
                 // indicating no change.
+                // In the case of a Create, null will be set and changed to the Lambda LogStreamName before sending.
                 client.send(event, context, Response.failed(event.getPhysicalResourceId()));
             } catch (Exception e) {
                 // unable to generate response AND send the failure
--- a/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/AbstractCustomResourceHandler.java
+++ b/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/AbstractCustomResourceHandler.java
@@ -65,7 +65,10 @@ public abstract class AbstractCustomResourceHandler
         } catch (CustomResourceResponseException rse) {
             LOG.error("Unable to generate response. Sending empty failure to {}", responseUrl, rse);
             try {
-                client.send(event, context, Response.failed());
+                // If the customers code throws an exception, Powertools should respond in a way that doesn't
+                // change the CloudFormation resources. A failure is sent with the existing PhysicalResourceId
+                // indicating no change.
+                client.send(event, context, Response.failed(event.getPhysicalResourceId()));
             } catch (Exception e) {
                 // unable to generate response AND send the failure
                 LOG.error("Unable to send failure response to {}.", responseUrl, e);
--- a/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
+++ b/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
@@ -56,20 +56,6 @@ class CloudFormationResponse {
         private final String logicalResourceId;
         private final boolean noEcho;
 
-        @Override
-        public String toString() {
-            final StringBuffer sb = new StringBuffer("ResponseBody{");
-            sb.append("status='").append(status).append('\'');
-            sb.append(", reason='").append(reason).append('\'');
-            sb.append(", physicalResourceId='").append(physicalResourceId).append('\'');
-            sb.append(", stackId='").append(stackId).append('\'');
-            sb.append(", requestId='").append(requestId).append('\'');
-            sb.append(", logicalResourceId='").append(logicalResourceId).append('\'');
-            sb.append(", noEcho=").append(noEcho);
-            sb.append('}');
-            return sb.toString();
-        }
-
         ResponseBody(CloudFormationCustomResourceEvent event,
                      Response.Status responseStatus,
                      String physicalResourceId,
@@ -128,6 +114,20 @@ class CloudFormationResponse {
                 node.set(DATA_PROPERTY_NAME, dataNode);
             }
             return node;
+        }
+
+        @Override
+        public String toString() {
+            final StringBuffer sb = new StringBuffer("ResponseBody{");
+            sb.append("status='").append(status).append('\'');
+            sb.append(", reason='").append(reason).append('\'');
+            sb.append(", physicalResourceId='").append(physicalResourceId).append('\'');
+            sb.append(", stackId='").append(stackId).append('\'');
+            sb.append(", requestId='").append(requestId).append('\'');
+            sb.append(", logicalResourceId='").append(logicalResourceId).append('\'');
+            sb.append(", noEcho=").append(noEcho);
+            sb.append('}');
+            return sb.toString();
         }
     }
 
@@ -212,6 +212,8 @@ class CloudFormationResponse {
 
     /**
      * Returns the response body as an input stream, for supplying with the HTTP request to the custom resource.
+     *
+     * If PhysicalResourceId is null at this point it will be replaced with the Lambda LogStreamName.
      *
      * @throws CustomResourceResponseException if unable to generate the response stream
      */
--- a/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
+++ b/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
@@ -226,7 +226,10 @@ class CloudFormationResponse {
                 ObjectNode node = body.toObjectNode(null);
                 return new StringInputStream(node.toString());
             } else {
-                String physicalResourceId = resp.getPhysicalResourceId() != null ? resp.getPhysicalResourceId() : context.getLogStreamName();
+
+                String physicalResourceId = resp.getPhysicalResourceId() != null ? resp.getPhysicalResourceId() :
+                        event.getPhysicalResourceId() != null? event.getPhysicalResourceId() : context.getLogStreamName();
+
                 ResponseBody body = new ResponseBody(event, resp.getStatus(), physicalResourceId, resp.isNoEcho(), reason);
                 LOG.debug("ResponseBody: {}", body);
                 ObjectNode node = body.toObjectNode(resp.getJsonNode());
--- a/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
+++ b/powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
@@ -53,14 +53,14 @@ class CloudFormationResponse {
         private final boolean noEcho;
 
         ResponseBody(CloudFormationCustomResourceEvent event,
-                     Context context,
                      Response.Status responseStatus,
                      String physicalResourceId,
-                     boolean noEcho) {
+                     boolean noEcho,
+                     String reason) {
             Objects.requireNonNull(event, "CloudFormationCustomResourceEvent cannot be null");
-            Objects.requireNonNull(context, "Context cannot be null");
-            this.physicalResourceId = physicalResourceId != null ? physicalResourceId : context.getLogStreamName();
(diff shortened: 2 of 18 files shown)


powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/Response.java
```diff
@@ -138,23 +138,69 @@ public static Builder builder() {
...
+
+    /**
+     * Creates a successful Response with no physicalResourceId set. Powertools will set the physicalResourceId to the
+     * Lambda LogStreamName
+     *
+     * The value returned for a PhysicalResourceId can change custom resource update operations. If the value returned
+     * is the same, it is considered a normal update. If the value returned is different, AWS CloudFormation recognizes
+     * the update as a replacement and sends a delete request to the old resource. For more information,
+     * see AWS::CloudFormation::CustomResource.
+     *
```
jeromevdl, line 180:
```suggestion
     *
     * @deprecated this method is not safe. Provide a physicalResourceId.
```

powertools-cloudformation/src/test/java/software/amazon/lambda/powertools/cloudformation/handlers/RuntimeExceptionThrownHandler.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @Override
+    protected Response update(CloudFormationCustomResourceEvent event, Context context) {
+        throw new RuntimeException("failure");
+    }
+
+    @Override
+    protected Response delete(CloudFormationCustomResourceEvent event, Context context) {
+        return null;
```
jeromevdl, line 22: what happens if there is an exception in the delete? do we have an IT for this case?

jeromevdl (changes requested):

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
```diff
@@ -146,7 +164,7 @@ SdkHttpClient getClient() {
      */
     public HttpExecuteResponse send(CloudFormationCustomResourceEvent event,
                                     Context context) throws IOException, CustomResourceResponseException {
-        return send(event, context, null);
+        return send(event, context, Response.success(context.getLogGroupName()));
```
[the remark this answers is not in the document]
scottgerring: @jeromevdl We talked about this a bit and we're getting a bit close to the "breaking change" line. Having said that, as we're only going to plug this into CREATE requests after this PR (where UPDATE/DELETE are echoed back from the CFN request if not supplied), it should only cause issues if clients are inferring meaning from the physicalResourceId returned - e.g. trying to parse it or something.

powertools-cloudformation/src/test/java/software/amazon/lambda/powertools/cloudformation/handlers/RuntimeExceptionThrownHandler.java
scottgerring: It _should_ be the same as for an UPDATE (echo back request phyId) but I don't think we have coverage. Is there a standard way of doing parameterized tests in Java these days - I think some of these tests would work quite nicely parameterized out and let us more easily understand the intent - here for instance you want to assert the behaviour is the same for UPDATE/DELETE for both exceptional and null responses.

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
```diff
@@ -32,6 +34,8 @@
  */
 class CloudFormationResponse {
 
+    private static final Logger LOG = LoggerFactory.getLogger(CloudFormationResponse.class);
```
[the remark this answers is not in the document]
jeromevdl: sorry

powertools-cloudformation/src/test/java/software/amazon/lambda/powertools/cloudformation/handlers/RuntimeExceptionThrownHandler.java
jeromevdl: https://junit.org/junit5/docs/current/api/org.junit.jupiter.params/org/junit/jupiter/params/ParameterizedTest.html

mriccia:
A further improvement on this functionality would be to allow Powertools users to specify which `physicalResourceId` to use, in the case of an exception. 
This may not be required for this PR, but worth considering in future improvements.

A few scenarios where this would be useful:

-  during an UPDATE, the customer code may be trying to create a new resource. if an exception is thrown, a FAILED response is sent to CloudFormation with the previous `physicalResourceId`. During the rollback, the stack will trigger an UPDATE with the previous Custom Resource properties. 

Example: a custom resource creates an S3 bucket and with a specified lifecycle rule for expiring objects.  
When an UPDATE happens with a different AWS S3 bucket name, the Custom resource will need to create a new S3 bucket with the new bucket name. 
if the Custom Resource throws an exception **after** the new bucket is created, the response to CF will be with the previous `physicalResourceId`. During the rollback, CF will invoke the custom resource with the previous bucket name and lifecycle rules, leaving the possibility of an orphaned S3 bucket.
Instead, if the user can specify a different `physicalResourceId` (because they are replacing the underlying resources), during the failure CloudFormation will be notified of this replacement and during the rollback will issue a DELETE of the new `physicalResourceId`.
 
-  during a CREATE of a custom resource, there is an exception in the Handler code. `physicalResourceId` will not be present in the CloudFormation Event, therefore the LogStreamName is returned as the `physicalResourceId`. During the rollback, Cloudformation will issue a DELETE with the new `physicalResourceId`. The user code may not know what to do with `physicalResourceId` as it does not give it any context

powertools-cloudformation/src/test/java/software/amazon/lambda/powertools/cloudformation/handlers/RuntimeExceptionThrownHandler.java
scottgerring: > Please can we add the following tests:
> 
>     * `CREATE` with a user provided `physicalResourceId` via `Response.success(physicalResourceId)` and `Response.failed(physicalResourceId)`- validate that the `physicalResourceId` is sent to CF
> 
>     * `UPDATE` with a user provided `physicalResourceId` via `Response.success(physicalResourceId)` and `Response.failed(physicalResourceId)`- validate that the `physicalResourceId` is sent to CF
> 
>     * `DELETE` with a user provided `physicalResourceId` via `Response.success(physicalResourceId)` and `Response.failed(physicalResourceId)`- validate that the `physicalResourceId` is sent to CF
> 
> 
> Adding these tests will ensure that we are indeed sending the provided `physicalResourceId` to Cloudformation, and they will help us catch any future regressions

Suggest doing this with a parameterized test per @jeromevdl 's link above; it'll make it easier to understand the intent - "C/U/D all pass the physicalResourceId along to CF on success"

scottgerring:
> Please can we add the following tests:
> 
>     * `CREATE` with a user provided `physicalResourceId` via `Response.success(physicalResourceId)` and `Response.failed(physicalResourceId)`- validate that the `physicalResourceId` is sent to CF
> 
>     * `UPDATE` with a user provided `physicalResourceId` via `Response.success(physicalResourceId)` and `Response.failed(physicalResourceId)`- validate that the `physicalResourceId` is sent to CF
> 
>     * `DELETE` with a user provided `physicalResourceId` via `Response.success(physicalResourceId)` and `Response.failed(physicalResourceId)`- validate that the `physicalResourceId` is sent to CF
> 
> 
> Adding these tests will ensure that we are indeed sending the provided `physicalResourceId` to Cloudformation, and they will help us catch any future regressions

@mriccia - Suggest doing this with a parameterized test per Jerome's link above; it'll make it easier to understand the intent - "C/U/D all pass the physicalResourceId along to CF on success"

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
```diff
@@ -195,23 +213,32 @@ protected Map<String, List<String>> headers(int contentLength) {
...
-                ResponseBody body = new ResponseBody(event, context, Response.Status.SUCCESS, null, false);
+                ResponseBody body = new ResponseBody(event, Response.Status.SUCCESS, context.getLogStreamName(), false, reason);
+                LOG.debug("ResponseBody: {}", body);
                 ObjectNode node = body.toObjectNode(null);
                 return new StringInputStream(node.toString());
             } else {
-                ResponseBody body = new ResponseBody(
-                        event, context, resp.getStatus(), resp.getPhysicalResourceId(), resp.isNoEcho());
+
+                String physicalResourceId = resp.getPhysicalResourceId() != null ? resp.getPhysicalResourceId() :
```
scottgerring, line 232: ID injection - and here

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/AbstractCustomResourceHandler.java
```diff
@@ -65,7 +65,12 @@ public final Response handleRequest(CloudFormationCustomResourceEvent event, Con
         } catch (CustomResourceResponseException rse) {
             LOG.error("Unable to generate response. Sending empty failure to {}", responseUrl, rse);
             try {
-                client.send(event, context, Response.failed());
+                // If the customers code throws an exception, Powertools should respond in a way that doesn't
+                // change the CloudFormation resources.
+                // In the case of a Update or Delete, a failure is sent with the existing PhysicalResourceId
+                // indicating no change.
+                // In the case of a Create, null will be set and changed to the Lambda LogStreamName before sending.
+                client.send(event, context, Response.failed(event.getPhysicalResourceId()));
```
scottgerring, line 73: ID Injection - we're doing this in two places - here, and further down the page-and-also-call-stack (comment added). On the one hand we're trying to keep the interfaces from changing at this point and deferring that to V2, on the other hand and we've got solid test coverage now, on the other, its a bit irky that this piece of fairly important logic is spread about.

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
scottgerring, line 167: physicalResourceId injection - and here

scottgerring (commented):
The implementation looks like it satisfies our goals of 1/ resolving #1081 while 2/ not introducing breaking changes; both extensive discussion and reviewing of the code and our new test coverage help ensure this.

I have one outstanding question inline about the logic for physicalResourceId choice being spread about. This might just how it has to be to maintain the interfaces, but I wanted to raise for discussion before we merge.

mriccia (commented):
Thanks for adding in the additional tests Scott.
We should make 1 modification before pushing, in CloudFormationResponse.java [here ](https://github.com/awslabs/aws-lambda-powertools-java/pull/1082/files#diff-a6063abba30fb90e067ad04e43a6243a0658d9ab449dfb779da13ae9dfcfb987R167) and [here](https://github.com/awslabs/aws-lambda-powertools-java/pull/1082/files#diff-a6063abba30fb90e067ad04e43a6243a0658d9ab449dfb779da13ae9dfcfb987R226) we should also retrieve the `physicalResourceId` from the event if present.

Separately, I agree that we need to centralise the logic on assigning `physicalResourceId` to a single place, in my view this can be done in V2.

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
```diff
@@ -195,23 +213,32 @@ protected Map<String, List<String>> headers(int contentLength) {
...
      * @throws CustomResourceResponseException if unable to generate the response stream
      */
     StringInputStream responseBodyStream(CloudFormationCustomResourceEvent event,
                                          Context context,
                                          Response resp) throws CustomResourceResponseException {
         try {
+            String reason = "See the details in CloudWatch Log Stream: " + context.getLogStreamName();
             if (resp == null) {
-                ResponseBody body = new ResponseBody(event, context, Response.Status.SUCCESS, null, false);
+                ResponseBody body = new ResponseBody(event, Response.Status.SUCCESS, context.getLogStreamName(), false, reason);
```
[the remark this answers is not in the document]
scottgerring: @mriccia this does smell a bit. I suppose it's being dealt with by the incoming calls handling it, else we'd have a test failure. Or are we missing a test?

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/CloudFormationResponse.java
scottgerring: @mriccia it's `public` - not sure we can change it in case a consumer of the powertools is using it?

powertools-cloudformation/src/main/java/software/amazon/lambda/powertools/cloudformation/AbstractCustomResourceHandler.java
msailes: I agree, closing.

scottgerring (approved):

mriccia (approved):
LGTM

jeromevdl (approved):

Issue #1083, opened 2023-03-10
Add DynamoDB provider for the parameters module


jeromevdl:
To fill the gap with Powertools for Python, we should add DynamoDB as a source for parameters.

Python version: https://awslabs.github.io/aws-lambda-powertools-python/2.9.1/utilities/parameters/#dynamodbprovider

In order to provide this, we can implement a new Provider (`DynamoDBProvider`) extending the [`BaseProvider`](https://github.com/awslabs/aws-lambda-powertools-java/blob/master/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java) we already have.

Note that adding this implies to add a dependency to the DynamoDB SDK, which will make the module heavier. Thinking about a v2, we could permit to choose only the required provider (and dependencies) to limit the size of the module.

scottgerring:
Implemented, in next release

PR #1091, opened 2023-03-16
[WIP feat] Add DynamoDB provider to parameters module

scottgerring:
**Issue #, if available:**
#1083 

## Description of changes:

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-parameters/pom.xml
+++ b/powertools-parameters/pom.xml
@@ -74,6 +74,10 @@
             <artifactId>url-connection-client</artifactId>
         </dependency>
         <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>dynamodb</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
--- a/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDBProvider.java
+++ b/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDBProvider.java
@@ -1,0 +1,69 @@
+package software.amazon.lambda.powertools.parameters;
+
+import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.*;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class DynamoDBProvider extends BaseProvider {
+
+    private final DynamoDbClient client;
+    private final String tableName;
+
+    public DynamoDBProvider(CacheManager cacheManager, String tableName) {
+        this(cacheManager, DynamoDbClient.builder()
+                .httpClientBuilder(UrlConnectionHttpClient.builder())
+                .credentialsProvider(EnvironmentVariableCredentialsProvider.create())
+                .region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))
+                .build(),
+                tableName
+        );
+
+    }
+
+    DynamoDBProvider(CacheManager cacheManager, DynamoDbClient client, String tableName) {
+        super(cacheManager);
+        this.client = client;
+        this.tableName = tableName;
+    }
+
+    @Override
+    protected String getValue(String key) {
+        GetItemResponse resp = client.getItem(GetItemRequest.builder()
+                .tableName(tableName)
+                .key(Collections.singletonMap("id", AttributeValue.fromS(key)))
+                .attributesToGet("val")
+                .build());
+
+        // TODO
+        return resp.item().values().stream().findFirst().get().s();
+    }
+
+    @Override
+    protected Map<String, String> getMultipleValues(String path) {
+
+        QueryResponse resp = client.query(QueryRequest.builder()
+                .tableName(tableName)
+                .keyConditionExpression("id = :v_id")
+                .expressionAttributeValues(Collections.singletonMap("v_id", AttributeValue.fromS(path)))
+                .attributesToGet("sk", "value")
+                .build());
+
+        return resp
+                    .items()
+                    .stream()
+                    .collect(
+                            Collectors.toMap(
+                                    (i) -> i.get("sk").s(),
+                                    (i) -> i.get("val").s()));
+
+
+    }
+}
--- a/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDBProvider.java
+++ b/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDBProvider.java
@@ -20,8 +20,8 @@ public class DynamoDBProvider extends BaseProvider {
     public DynamoDBProvider(CacheManager cacheManager, String tableName) {
         this(cacheManager, DynamoDbClient.builder()
                 .httpClientBuilder(UrlConnectionHttpClient.builder())
-                .credentialsProvider(EnvironmentVariableCredentialsProvider.create())
-                .region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))
+                //.credentialsProvider(EnvironmentVariableCredentialsProvider.create())
+                //.region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))
                 .build(),
                 tableName
         );
@@ -42,8 +42,14 @@ public class DynamoDBProvider extends BaseProvider {
                 .attributesToGet("val")
                 .build());
 
-        // TODO
-        return resp.item().values().stream().findFirst().get().s();
+        // If we have an item at the key, we should be able to get a 'val' out of it. If not it's
+        // exceptional.
+        // If we don't have an item at the key, we should return null.
+        if (resp.hasItem() && !resp.item().values().isEmpty()) {
+            return resp.item().values().stream().findFirst().get().s();
+        }
+
+        return null;
     }
 
     @Override
@@ -52,8 +58,7 @@ public class DynamoDBProvider extends BaseProvider {
         QueryResponse resp = client.query(QueryRequest.builder()
                 .tableName(tableName)
                 .keyConditionExpression("id = :v_id")
-                .expressionAttributeValues(Collections.singletonMap("v_id", AttributeValue.fromS(path)))
-                .attributesToGet("sk", "value")
+                .expressionAttributeValues(Collections.singletonMap(":v_id", AttributeValue.fromS(path)))
                 .build());
 
         return resp
--- a/powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDBProviderE2ETest.java
+++ b/powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDBProviderE2ETest.java
@@ -1,0 +1,89 @@
+package software.amazon.lambda.powertools.parameters;
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.services.dynamodb.model.PutItemRequest;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * This class provides simple end-to-end style testing of the DynamoDBProvider class.
+ * It is ignored, for now, as it requires AWS access and that's not yet run as part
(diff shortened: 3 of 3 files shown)


powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDBProvider.java
```diff
@@ -0,0 +1,74 @@
...
+            return resp.item().values().stream().findFirst().get().s();
+        }
+
+        return null;
+    }
+
+    @Override
+    protected Map<String, String> getMultipleValues(String path) {
+
+        QueryResponse resp = client.query(QueryRequest.builder()
```
stevehouel, line 58: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This code might not produce accurate results if the operation returns paginated results instead of all results. Consider adding another call to check for additional results.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDBProvider.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.stream.Collectors;
+
+public class DynamoDBProvider extends BaseProvider {
+
+    private final DynamoDbClient client;
+    private final String tableName;
+
+    public DynamoDBProvider(CacheManager cacheManager, String tableName) {
+        this(cacheManager, DynamoDbClient.builder()
+                .httpClientBuilder(UrlConnectionHttpClient.builder())
```
scottgerring, line 22: @jeromevdl , is there any reason that this code in e.g. the SecretsProvider explicitly configures the AWS SDK client? My experience has been that the default configuration chain is usually a safe bet - am I missing something?

https://github.com/awslabs/aws-lambda-powertools-java/blob/b062a7622d6845e1ea775a7dcd7c3658315e6dfa/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/SecretsProvider.java#L66-L72

jeromevdl: Actually I don't know/remember. But it works, `AWS_ACCESS_KEY`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN` are available: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

jeromevdl: For the http client, we could now move to CRT (https://aws.amazon.com/blogs/developer/announcing-availability-of-the-aws-crt-http-client-in-the-aws-sdk-for-java-2-x/).

msailes: `.credentialsProvider(EnvironmentVariableCredentialsProvider.create())` was definitely the best way to configure clients pre SnapStart. Now it's probably best to use the default.

jeromevdl: Yep, I remember now. By specifying environment variables, the SDK goes directly that way and doesn't look for multiple sources. It's faster. Not sure everyone is using snap start though. Unless it is a problem for snap start (is it?), I would keep it like that.

scottgerring: Thanks both! I'll plug in CRT and use the env vars. It's nice if it generalizes outside of Lambda, but our explicit goal here is to run in Lambda, so it's a bit unnecessary.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDbProvider.java
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Returns multiple values from the DynamoDB parameter provider.
+     *
+     * @param path Parameter store path
+     * @return All values matching the given path, and an empty map if none do. Throws if any records exist that don't match the schema.
+     */
+    @Override
+    protected Map<String, String> getMultipleValues(String path) {
+
+        QueryResponse resp = client.query(QueryRequest.builder()
```
stevehouel, line 78: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This code might not produce accurate results if the operation returns paginated results instead of all results. Consider adding another call to check for additional results.

powertools-parameters/pom.xml
```diff
@@ -73,6 +73,10 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>url-connection-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
```
scottgerring, line 77: I have taken this as a dep on the powertools-parameters project. I considered refactoring into submodules so you don't take the dynamodb dependency if you use parameters at all, but agree with @jeromevdl that this is better for a v2 change, as we'll have to move the existing providers down too and this will break.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/ParamManagerIntegrationTest.java
```diff
@@ -116,4 +120,16 @@ public void secretsProvider_get() {
         assertThat(secretsProvider.get("keys")).isEqualTo(expectedValue); // second time is from cache
         verify(secretsManagerClient, times(1)).getSecretValue(any(GetSecretValueRequest.class));
     }
+
+    @Test
+    public void getDynamoDbProvider() {
```
scottgerring, line 125: Here I simply test we can get a provider through the `ParamManager`. Usage of the provider is tested in `DynamoDbProviderTest`.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+        // Act
+        Map<String, String> values = provider.getMultipleValues(UUID.randomUUID().toString());
+
+        // Assert
+        assertThat(values.size()).isEqualTo(0);
+    }
+
+    @Test
+    public void getValuesWithMalformedRowThrows() {
```
scottgerring, line 147: It's not documented on the base interface what happens if you ask for something and the data that comes back is either 1/ not there or 2/ doesn't match our schema. I am asserting that for 1/ we return null and 2/ we throw.

docs/utilities/parameters.md
```diff
@@ -166,6 +168,52 @@ in order to get data from other regions or use specific credentials.
...
+    public class AppWithDynamoDbParameters implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+        // Get an instance of the DynamoDbProvider
+        DynamoDbProvider ddbProvider = ParamManager.getDynamoDbProvider("my-parameters-table");
+    
+        // Retrieve a single parameter
+        String value = ddbProvider.get("my-key"); 
+    } 
+    ```
+
+=== "DynamoDbProvider with an explicit region"
```
jeromevdl, line 191: maybe say "custom dynamodb client" rather than "explicit region"

powertools-parameters/pom.xml
jeromevdl: yep, it will burden the jar, we'll split it in v2

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDbProvider.java
```diff
@@ -0,0 +1,179 @@
...
+                .httpClientBuilder(UrlConnectionHttpClient.builder())
+                .credentialsProvider(EnvironmentVariableCredentialsProvider.create())
+                .region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))
+                .build(),
+                tableName
+        );
+
+    }
+
+    DynamoDbProvider(CacheManager cacheManager, DynamoDbClient client, String tableName) {
```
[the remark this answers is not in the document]
jeromevdl: ok

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderE2ETest.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * This class provides simple end-to-end style testing of the DynamoDBProvider class.
+ * It is ignored, for now, as it requires AWS access and that's not yet run as part
+ * of our unit test suite in the cloud.
+ *
+ * The test is kept here for 1/ local development and 2/ in preparation for future
+ * E2E tests running in the cloud CI.
+ */
+@Disabled
+public class DynamoDbProviderE2ETest {
```
jeromevdl, line 25: Should we have this in the e2e tests module instead? You can look in the idempotency module for tests with DDB: https://github.com/awslabs/aws-lambda-powertools-java/blob/master/powertools-idempotency/src/test/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStoreTest.java

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Test
+    public void getValue() {
+
+        // Arrange
+        String key = "Key1";
+        String expectedValue = "Value1";
+        HashMap<String, AttributeValue> responseData = new HashMap<String, AttributeValue>();
+        responseData.put("id", AttributeValue.fromS(key));
+        responseData.put("val", AttributeValue.fromS(expectedValue));
```
jeromevdl, line 52: we should use "value" instead, to be compatible with python: https://awslabs.github.io/aws-lambda-powertools-python/2.9.1/utilities/parameters/#dynamodbprovider

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+
+    @Test
+    public void getValue() {
+
+        // Arrange
+        String key = "Key1";
+        String expectedValue = "Value1";
+        HashMap<String, AttributeValue> responseData = new HashMap<String, AttributeValue>();
+        responseData.put("id", AttributeValue.fromS(key));
```
jeromevdl, line 51: I don't like "id", but this is what they use in python...

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderTest.java
```diff
@@ -0,0 +1,167 @@
...
+        // Arrange
+        String key = "Key1";
+        HashMap<String, AttributeValue> responseData = new HashMap<String, AttributeValue>();
+        responseData.put("id", AttributeValue.fromS(key));
+        responseData.put("not-val", AttributeValue.fromS("something"));
+        Mockito.when(client.getItem(getItemValueCaptor.capture())).thenReturn(GetItemResponse.builder()
+                .item(responseData)
+                .build());
+        // Act
+        Assertions.assertThrows(NullPointerException.class, () -> {
```
jeromevdl, line 93: Are you sure it should be nullpointer? Couldn't we have a better (eventually custom) one?

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderTest.java
jeromevdl: agree on that, but throw something else than nullpointer

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderE2ETest.java
scottgerring: Agreed! As it's not there yet I'd intended to merge it in with this commit so its complete, and then move it over after the E2E one is in master. WDYT @jeromevdl ?

jeromevdl: OK to keep it here while we merge the e2e in master. But we'll need to add the CDK part to it, to make it autonomous.

docs/utilities/parameters.md
scottgerring: I like this better - it makes it clear that it let's you do more than just customize the region. I will update the other param provider comments accordingly.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderE2ETest.java
scottgerring: Agreed - so it'll use the E2E stuff to get itself its own DDB tables for each schema, seed them with data, then test against them - rather than hardcoding some tables that the developer happens to have floating around in their AWS account.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/DynamoDbProviderTest.java
scottgerring: Throwing a `DynamoDbProviderSchemaException` now with the error text describing the missing column

scottgerring:
@jeromevdl I think that should be it! I've cleaned up the exceptions too.

scottgerring:
@jeromevdl anything else to polish? 👼

jeromevdl (approved):
LGTM

Issue #1080, opened 2023-03-03
Maintenance: Add unit test coverage


jeromevdl:
### Summary

Add unit test coverage to get a better visibility on where we can have issues, and improve the coverage where needed.

### Why is this needed?

Lambda Powertools for Java has a decent amount of unit tests but we don't have a clear view on what is well-tested and more important where we can improve. Having the coverage reports will help driving the effort and make sure we minimize the risk of issues.

### Which area does this relate to?

Tests

### Solution

We could use one of the standard frameworks (Jacoco, Cobertura) and deploy the reports to codecov.io just like the Python version does (see https://github.com/awslabs/aws-lambda-powertools-python/blob/79e7b4392a1a0f69bcf627b137c81be45867c587/.github/workflows/python_build.yml#L55)

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-java/#tenets)
- [X] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

scottgerring:
The [powertools for python](https://app.codecov.io/gh/awslabs/aws-lambda-powertools-python) dashboard

jeromevdl:
closed with #1094

PR #1094, opened 2023-03-17
maintenance: send coverage (jacoco) to codecov

jeromevdl:
**Issue #, if available:** #1080

## Description of changes:

- removed cobertura as we have also jacoco
- merge results in parent target folder

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,6 +64,8 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70 # 3.1.1
         if: ${{ matrix.java == '11.0.x' }} # publish results once
+        with:
+          directory: ./target
   savepr:
     runs-on: ubuntu-latest
     name: Save PR number if running on PR by dependabot
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,6 @@ jobs:
         java: [8, 8.0.192, 11.0.x, 11.0.3, 12, 13, 15, 16, 17 ]
     name: Java ${{ matrix.java }}
     env:
-      OS: ${{ matrix.os }}
       JAVA: ${{ matrix.java }}
       AWS_REGION: eu-west-1
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
         uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70 # 3.1.1
         if: ${{ matrix.java == '11.0.x' }} # publish results once
         with:
-          directory: ./target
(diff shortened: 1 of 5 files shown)


jeromevdl:
It works: https://app.codecov.io/github/awslabs/aws-lambda-powertools-java/tree/unittest_coverage_report

examples/powertools-examples-validation/pom.xml
```diff
@@ -6,12 +6,11 @@
     <name>AWS Lambda Powertools for Java library Examples - Validation</name>
 
     <parent>
-        <artifactId>powertools-parent</artifactId>
+        <artifactId>powertools-examples</artifactId>
```
scottgerring, line 9: 🤭 thanks for the cleanup

.github/workflows/build.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     name: Java ${{ matrix.java }}
     env:
       OS: ${{ matrix.os }}
-      JAVA: ${{ matrix.java-version }}
+      JAVA: ${{ matrix.java }}
```
scottgerring, line 53: It's not clear to me what the previous version of this was doing ..... shouldn't `matrix.java-version` have always been null? :|

jeromevdl: I think so. Not sure the environment variable is even used actually...

.github/workflows/build.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     name: Java ${{ matrix.java }}
     env:
       OS: ${{ matrix.os }}
```
scottgerring, line 52: Only related to change insofar as `matrix.java-version` was also not defined in the matrix context. In this case I think `matrix.os` has been taken from the [matrix examples](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) but is being misused here because it is never set.

I suggest removing this. I can see no usages of `OS` within the codebase, and this is actively confusing. Adding the [filter on the matrix to ensure codecov runs once](https://github.com/awslabs/aws-lambda-powertools-java/pull/1094/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R67) adds to this - if we actually had a 2D matrix with OS as well, which this implies, the filter would not be enough.

.github/workflows/build.yml
```diff
@@ -59,8 +59,14 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: ${{ matrix.java }}
+          cache: 'maven'
       - name: Build with Maven
         run: mvn -Pbuild-without-spotbugs -B package --file pom.xml
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70 # 3.1.1
+        if: ${{ matrix.java == '11.0.x' }} # publish results once
```
scottgerring, line 67: This is good as long as we have a 1D matrix only. We could also use [`jobs.<job_id>.needs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) to structure this more explicitly as a fan-in, but I think this would make the build slower because that step would then have to rebuild the codebase and re-run the tests to get coverage. 

I think we should keep it this way because end-to-end build time is important, we just have to remember that if we ever want to do builds along another axis we're going to have to restructure this

scottgerring (changes requested):
A couple comments inline that are tenuously related about the way we're using the build matrix

.github/workflows/build.yml
jeromevdl: I'm not an expert in GitHub actions. Happy to replace this by something else better. What I wanted is to ensure we don't try to publish codecov results in all builds.

.github/workflows/build.yml
jeromevdl: yep, will try to remove the os to see if it still works.

.github/workflows/build.yml
scottgerring: I think the way it is now is best because if we make it more "general" it'll be slower (re-run build and tests) for no benefit at this point because we only matrix on one axis !

jeromevdl:
@scottgerring anything else?

scottgerring (approved):
LGTM!

Issue #1084, opened 2023-03-10
Add AppConfig provider for the parameters module


jeromevdl:
To fill the gap with Powertools for Python, we should add AppConfig as a source for parameters.

Python version: https://awslabs.github.io/aws-lambda-powertools-python/2.9.1/utilities/parameters/#appconfigprovider

In order to provide this, we can implement a new Provider (`AppConfigProvider`) extending the [`BaseProvider`](https://github.com/awslabs/aws-lambda-powertools-java/blob/master/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java) we already have.

Note that adding this implies to add a dependency to the AppConfig SDK, which will make the module heavier. Thinking about a v2, we could permit to choose only the required provider (and dependencies) to limit the size of the module.

Probably in v2, if we separate parameters in submodules and have one for AppConfig, it could also be used for Feature Flag module (if reuse is possible and makes sense).

scottgerring:
I've started #1083 and it looks like it'll save time to do both of these once, so i'll pick this up too.

scottgerring:
#1083 has been implemented. I will use the experience there to do this one as a separate PR now.

scottgerring:
Merged; waiting for release.

scottgerring:
Released in 1.16.0

PR #1104, opened 2023-03-20
feat: Add AppConfig provider to parameters module

scottgerring:
**Issue #, if available:** #1084

## Description of changes:
Implements a parameters provider that uses AWS' AppConfig service as the backing store. 

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:


powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
```diff
@@ -0,0 +1,212 @@
...
+         * @param environment the AppConfig environment
+         * @return the builder to chain calls (eg. <pre>builder.withCacheManager().build()</pre>)
+         */
+        public AppConfigProvider.Builder withEnvironment(String environment) {
+            this.environment = environment;
+            return this;
+        }
+
+        /**
+         * <b>Mandatory</b>. Provide a CacheManager to the {@link AppConfigProvider}
```
jeromevdl, line 180: Error in the comment...

powertools-parameters/pom.xml
```diff
@@ -77,6 +77,11 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>dynamodb</artifactId>
         </dependency>
+        <dependency>
```
jeromevdl, line 80: General comment, not to solve here: We're adding a new dependency, making the parameters module heavier (we did it already with DynamoDB). 

Should we think about different parameter modules (one per persistence technology: param store, DDB, app config, secrets manager) to avoid a big one with everything inside?

Note: parameter module is 29KB in v1.14, 34KB in v1.15 (with the add of DDB), how much with AppConfig?
29 or 39kb is not a big deal but just to discuss...

jeromevdl: @msailes your thoughts on this?

jeromevdl (commented):
Good job, just a few comments.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
```diff
@@ -0,0 +1,212 @@
...
+    AppConfigProvider(CacheManager cacheManager, AppConfigDataClient client, String environment, String application) {
+        super(cacheManager);
+        this.client = client;
+        this.application = application;
+        this.environment = environment;
+    }
+
+
+    @Override
+    protected String getValue(String key) {
```
leandrodamascena, line 69: Hello @scottgerring! I might suggest renaming the name of this parameter to `profile` instead of `key`. 

AppConfig uses the idea of `profile`  which can be a JSON/TXT/YAML or even a different source like S3.

powertools-parameters/pom.xml
msailes: Given the variety of sources that could be introduced, I think this makes sense for a v2 ticket.

scottgerring: My feeling is we should nest modules again. The parameters modules should all standalone (apart from the base interface pieces of course); may as well let people pull them in separately. ACK that it feels a bit much for this change (and would be interface breaking for existing params providers)

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
```diff
@@ -0,0 +1,212 @@
...
+    private final String application;
+
+    private final String environment;
+
+    private final HashMap<String, EstablishedSession> establishedSessions = new HashMap<>();
+
+    public AppConfigProvider(CacheManager cacheManager, String environment, String application) {
+        this(cacheManager, AppConfigDataClient.builder()
+                        .httpClientBuilder(UrlConnectionHttpClient.builder())
+                        .credentialsProvider(EnvironmentVariableCredentialsProvider.create())
```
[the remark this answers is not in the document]
scottgerring: Hey @jeromevdl we've still got this scattered around everywhere (e.g. all the other persistence providers). I can fix it here but i'd say we want some common helper for it probably in `powertools-core`; the code to resolve the right provider is a bit much to copy-pasta:

```java
             // AWS_LAMBDA_INITIALIZATION_TYPE has two values on-demand and snap-start
                // when using snap-start mode, the env var creds provider isn't used and causes a fatal error if set
                // fall back to the default provider chain if the mode is anything other than on-demand.
                String initializationType = System.getenv().get(AWS_LAMBDA_INITIALIZATION_TYPE);
                if (initializationType  != null && initializationType.equals(ON_DEMAND)) {
                    ddbBuilder.credentialsProvider(EnvironmentVariableCredentialsProvider.create());
                }
```

WDYT?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
scottgerring: Hey @leandrodamascena I went to change this and then realized it is the name of the field in the base class - `    protected abstract String getValue(String key);`. Because it's used in terms of the common interface I've added javadoc to the method to explain that it maps back to an AppConfig `profile` in this case

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/testutils/Infrastructure.java
```diff
@@ -275,9 +290,73 @@ private Stack createStackWithLambda() {
                     .tableName(idempotencyTable)
                     .timeToLiveAttribute("expiration")
                     .build();
+
             table.grantReadWriteData(function);
         }
 
+        if (appConfig != null) {
```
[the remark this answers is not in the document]
scottgerring: how do you mean? I started with separate fields for each on the infra builder, but you end up with lots of null checking - if you've given me configurationValues, then you need an application and environment, if you've given me an application, you need an enviroment, etc. etc. I think the intent here is that it is all non-nullable and wrapping it up into a POJO is the best I can do for that. I welcome any other suggestions!

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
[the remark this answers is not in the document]
scottgerring: @jeromevdl @msailes i've changed this in the same fashion as the other providers - pushing the client creation into the builder only, and selecting a credentials provider based on snapstart. Can you have a quick sanity check?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
```diff
@@ -0,0 +1,226 @@
...
+                AppConfigDataClientBuilder ssmClientBuilder = AppConfigDataClient.builder()
+                        .httpClientBuilder(UrlConnectionHttpClient.builder())
+                        .region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())));
+
+                // AWS_LAMBDA_INITIALIZATION_TYPE has two values on-demand and snap-start
+                // when using snap-start mode, the env var creds provider isn't used and causes a fatal error if set
+                // fall back to the default provider chain if the mode is anything other than on-demand.
+                String initializationType = System.getenv().get(AWS_LAMBDA_INITIALIZATION_TYPE);
+                if (initializationType  != null && initializationType.equals(LambdaConstants.ON_DEMAND)) {
+                    ssmClientBuilder.credentialsProvider(EnvironmentVariableCredentialsProvider.create());
```
jeromevdl, line 158: Aren't you missing something? `ssmClientBuilder.build()` is not called, client cannot be initialised...

jeromevdl (commented):
please fix that one ;)

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
scottgerring: Fixed! And this is why we do code reviews 🤣  🤦

jeromevdl (commented):
LGTM

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/testutils/Infrastructure.java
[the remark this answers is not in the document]
jeromevdl: Just saying that's a lot of code to retrieve parameters, but it's just how AppConfig works...

scottgerring:
I reckon this is good to go. Happy to approve @jeromevdl ?

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
```diff
@@ -0,0 +1,228 @@
...
+        String nextSessionToken = response.nextPollConfigurationToken();
+
+        // Get the value of the key. Note that AppConfig will return null if the value
+        // has not changed since we last asked for it in this session - in this case
+        // we return the value we stashed at last request.
+        String value = response.configuration() != null?
+                response.configuration().asUtf8String() : // if we have a new value, use it
+                    establishedSession != null?
+                            establishedSession.lastConfigurationValue : // if we don't but we have a previous value, use that
+                            null; // otherwise we've got no value
```
kozub, line 101: How about using `if` statement instead of using nested ternary operator? IMHO it might be a bit more readable: 

```
        String value = null;
        if(response.configuration() != null) { // if we have a new value, use it
            value = response.configuration().asUtf8String();
        } else if (establishedSession != null){ // if we don't but we have a previous value, use that
            value = establishedSession.lastConfigurationValue;
        }
```

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
```diff
@@ -0,0 +1,228 @@
...
+        // Update the cache so we can get the next value later
+        establishedSessions.put(key, new EstablishedSession(nextSessionToken, value));
+
+        return value;
+    }
+
+    @Override
+    protected Map<String, String> getMultipleValues(String path) {
+        // Retrieving multiple values is not supported with the AppConfig provider.
+        throw new RuntimeException("Not implemented");
```
kozub, line 112: I would suggest changing exception message to add information that this operation is not supported by AppConfigProvider.

powertools-parameters/src/test/java/software/amazon/lambda/powertools/parameters/AppConfigProviderTest.java
```diff
@@ -0,0 +1,152 @@
...
+    private AppConfigProvider provider;
+
+    @Captor
+    ArgumentCaptor<StartConfigurationSessionRequest> startSessionRequestCaptor;
+
+    @Captor
+    ArgumentCaptor<GetLatestConfigurationRequest> getLatestConfigurationRequestCaptor;
+
+    @Captor
+    ArgumentCaptor<String> getValueForFieldCaptor;
```
kozub, line 41: Not used :)

powertools-e2e-tests/handlers/parameters/src/main/java/software/amazon/lambda/powertools/e2e/Input.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Map;
+
+public class Input {
+
+    private String app;
+    private String environment;
+    private String key;
+
+    public Input() {
```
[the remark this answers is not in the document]
scottgerring: Nope! 👍

powertools-e2e-tests/handlers/parameters/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+
+    <parent>
+        <groupId>software.amazon.lambda</groupId>
+        <artifactId>e2e-test-handlers-parent</artifactId>
+        <version>1.0.0</version>
+    </parent>
+
+    <artifactId>e2e-test-handler-parameters</artifactId>
+    <packaging>jar</packaging>
+    <name>A Lambda function using powertools logging</name>
```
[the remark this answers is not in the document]
scottgerring: 👍  The other modules need this change too, but I don't want to mix that into this PR, so we can do it separately.

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
scottgerring: It is a big assignment, but I personally find it easier to reason about this way - you assign the variable once in one place and it's easier to refactor for that - you can't have more code creep in between variable definition and assignment of value that might break something. The alternate would also need another branch too - `establishedSession` is also nullable.

jeromevdl (approved):

jeromevdl (approved):

PR #1175, opened 2023-06-04
chore: E2E tests GitHub action

jeromevdl:
**Issue #, if available:**

## Description of changes:

Run end-to-end tests with GitHub actions. 

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -12,6 +12,7 @@ on:
       - 'powertools-logging/**'
       - 'powertools-tracing/**'
       - 'powertools-idempotency/**'
+      - 'powertools-parameters/**'
       - 'powertools-metrics/**'
       - 'pom.xml'
       - '.github/workflows/**'
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -44,4 +44,4 @@ jobs:
           role-to-assume: ${{ secrets.AWS_ROLE_ARN_TO_ASSUME }}
           aws-region: ${{ env.AWS_DEFAULT_REGION }}
       - name: Run e2e test with Maven
-        run: mvn -Pe2e -B verify --file powertools-e2e-tests/pom.xml
+        run: export JAVA_VERSION=${{ matrix.java }} && mvn -Pe2e -B verify --file powertools-e2e-tests/pom.xml
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -20,9 +20,9 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 2
+      max-parallel: 3
       matrix:
-        java: [ 8, 11 ]
+        java: [ 8, 11, 17 ]
     name: End-to-end tests java${{ matrix.java }}
     env:
       JAVA_VERSION: ${{ matrix.java }}
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -1,0 +1,47 @@
+name: Run end-to-end tests
+
+on:
+  workflow_dispatch:
+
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -286,7 +286,7 @@
             <dependency>
                 <groupId>ch.qos.logback</groupId>
                 <artifactId>logback-classic</artifactId>
-                <version>1.4.7</version>
+                <version>1.3.4</version>
```
mriccia, line 289: Why the version downgrade?

jeromevdl: because of this: https://github.com/awslabs/aws-lambda-powertools-java/actions/runs/5165759332/jobs/9305470825#step:5:382

mriccia: Got it. Seems that 1.3.X should support Java 8. Any issues with the 1.3.7?
https://mvnrepository.com/artifact/ch.qos.logback/logback-classic/1.3.7

jeromevdl: right... for some reason WebStorm proposed the 1.3.4... Just updated with 1.3.7

.github/workflows/run-e2e-tests.yml
```diff
@@ -0,0 +1,47 @@
+name: Run end-to-end tests
+
+on:
+  workflow_dispatch:
+
+  push:
+    branches: [master]
+    paths: # add other modules when there are under e2e tests
```
scottgerring, line 8: Can we add `powertools-parameters` now so that it'll pick up the tests i've added as well once these both merge 👼 ? I don't think a glob on a directory that doesn't exist should break

.github/workflows/run-e2e-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+      - 'powertools-serialization/**'
+      - 'powertools-logging/**'
+      - 'powertools-tracing/**'
+      - 'powertools-idempotency/**'
+      - 'powertools-metrics/**'
+      - 'pom.xml'
+      - '.github/workflows/**'
+
+jobs:
+  e2e:
```
scottgerring, line 20: Is there a way to ensure that `cdk destroy` always runs after regardless of what happens with the main process? It is easy to imagine that we quickly end up in a situation where something like github action worker timeouts periodically leave us in a state where we are cleaning up the AWS account by hand periodically.

jeromevdl: Discussed with Ruben, they don't have any other belt around GitHub potential issue, just around their own code, just like this AfterAll...

.github/workflows/run-e2e-tests.yml
```diff
@@ -0,0 +1,48 @@
+name: Run end-to-end tests
+
+on:
+  workflow_dispatch:
+
+  push:
+    branches: [master]
```
jeromevdl, line 7:
```suggestion
    branches: [main]
```

pom.xml
```diff
@@ -286,7 +286,7 @@
             <dependency>
                 <groupId>ch.qos.logback</groupId>
                 <artifactId>logback-classic</artifactId>
-                <version>1.4.7</version>
+                <version>1.3.7</version>
```
kozub, line 289: Confirmed with @jeromevdl that the change as 1.4.x is not compatible with Java 8. 
I suggest changing it to latest 1.3.x version (1.3.8)

jeromevdl, line 289:
```suggestion
                <version>1.3.8</version>
```

jeromevdl: done

powertools-e2e-tests/pom.xml
```diff
@@ -29,7 +29,7 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>1.4.7</version>
+            <version>1.3.7</version>
```
jeromevdl, line 32:
```suggestion
            <version>1.3.8</version>
```

kozub (approved):

Issue #1140, opened 2023-04-25
Improve handling of messages with same MessageGroupId in SQS Batch processor


MartinMitro:
## Key information

* RFC PR: (leave this empty)
* Related issue(s), if known:
* Area: (i.e. Tracer, Metrics, Logger, etc.)
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-java/#tenets): (Yes/no)

## Summary
[summary]: #summary

Once the first message of Batch fails and we know it will be returned back to queue, rest of the messages with same message group id should no be processed 

## Motivation
[motivation]: #motivation

We want to be complaint with the FIFO SQS principle and process messages within same group in order.

## Proposal
[proposal]: #proposal

If the processing of the batch message fails with retryable exception and message attributes contain message group id, then the id should be stored and subsequent message of batch should be checked against it. If there is a match, message processing should be skipped and message also returned to queue. 

## Drawbacks
[drawbacks]: #drawbacks

As in FIFO queues, next messages of the same message group will be processed only when the failing was in removed from queue.

## Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

* **What other designs have been considered? Why not them?**

* **What is the impact of not doing this?**

Messages are not processed in order. Example: Lets say we have entity, which with id 1, and that ID is also message group ID. Now we have stack of updates in FIFO queue, first one updates value of entity to A, second one to value B. First one fails lets say on connection problem and will be retried later, second one succeeds and value is updated to B. Once the visibility timeout of first message expires, it will be processed again and value of entity with id 1 will be updated to A. Entity ends with invalid state.

## Unresolved questions
[unresolved-questions]: #unresolved-questions

scottgerring:
Hey @MartinMitro thanks for raising this!

Looking to the [python powertools for inspiration](https://awslabs.github.io/aws-lambda-powertools-python/2.15.0/utilities/batch/#fifo-queues), I see:

> When using [SQS FIFO queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html), we will stop processing messages after the first failure, and return all failed and unprocessed messages in batchItemFailures. This helps preserve the ordering of messages in your queue.

[...]

PR #1183, opened 2023-06-08
[WIP] Handle batch failures in FIFO queues correctly

scottgerring:
**Issue #, if available:** #1140 

## Description of changes:
When processing a FIFO-queue batch, once a message for a given `messageGroupId` fails, [no subsequent messages in the group should be processed](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting). This ensures ordering of messages is maintained. 

An open question remains on the implementation (fail the whole batch or only the whole message group?) that must be resolved before this implementation can be completed.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
+++ b/powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
@@ -43,6 +43,9 @@ public final class SqsUtils {
     private static SqsClient client;
     private static S3Client s3Client;
 
+    // The attribute on an SQS-FIFO message used to record the message group ID
+    private static final String MessageGroupIdKey = "MessageGroupId";
+
     private SqsUtils() {
     }
 
@@ -491,12 +494,28 @@ public final class SqsUtils {
 
         BatchContext batchContext = new BatchContext(client);
 
+        List<String> failedMessageGroupIds = new ArrayList<>(); // Track which message groups have failed
         for (SQSMessage message : event.getRecords()) {
+
+            // If we are trying to process a message that has a messageGroupId, we are on a FIFO queue,
+            // and should not process it if a previous message in the same group in this batch has failed.
+            String messageGroupId = message.getAttributes() != null?
+                    message.getAttributes().get(MessageGroupIdKey) : null;
+
             try {
-                handlerReturn.add(handler.process(message));
-                batchContext.addSuccess(message);
+                if (messageGroupId != null && failedMessageGroupIds.contains(messageGroupId)) {
+                    batchContext.addFailure(message, null); // TODO - is this sensible
+                } else {
+                    handlerReturn.add(handler.process(message));
+                    batchContext.addSuccess(message);
+                }
             } catch (Exception e) {
                 batchContext.addFailure(message, e);
+
+                // Record that this messageGroupId has a failure
+                if (messageGroupId != null) {
+                    failedMessageGroupIds.add(messageGroupId);
+                }
                 LOG.error("Encountered issue processing message: {}", message.getMessageId(), e);
             }
         }


powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
```diff
@@ -43,6 +44,9 @@ public final class SqsUtils {
     private static SqsClient client;
     private static S3Client s3Client;
 
+    // The attribute on an SQS-FIFO message used to record the message group ID
+    private static final String MessageGroupIdKey = "MessageGroupId";
```
jeromevdl, line 48:
```suggestion
    private static final String MESSAGE_GROUP_ID_KEY = "MessageGroupId";
```

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
```diff
@@ -489,15 +493,41 @@ public static <R> List<R> batchProcessor(final SQSEvent event,
...
+                    batchContext.addSuccess(message);
+                } catch(Exception e){
+
+                    // Record the failure
+                    batchContext.addFailure(message, e);
+
+                    // If we are trying to process a message that has a messageGroupId, we are on a FIFO queue. A failure
+                    // now stops us from processing the rest of the batch.
+                    String messageGroupId = message.getAttributes() != null ?
+                            message.getAttributes().get(MessageGroupIdKey) : null;
```
jeromevdl, line 522:
```suggestion
                            message.getAttributes().get(MESSAGE_GROUP_ID_KEY) : null;
```

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/exception/SkippedMessageDueToFailedBatchException.java
```diff
@@ -0,0 +1,11 @@
+package software.amazon.lambda.powertools.sqs.exception;
+
+/**
+ * Thrown to indicate that a
```
jeromevdl, line 4: ... ?

powertools-sqs/src/test/java/software/amazon/lambda/powertools/sqs/SqsUtilsFifoBatchProcessorTest.java
```diff
@@ -0,0 +1,122 @@
...
+    private MockitoSession session;
+
+    @Mock
+    private SqsClient sqsClient;
+
+    @Captor
+    private ArgumentCaptor<DeleteMessageBatchRequest> deleteMessageBatchCaptor;
+
+    public SqsUtilsFifoBatchProcessorTest() throws IOException {
+        sqsBatchEvent  =  MAPPER.readValue(SqsUtilsFifoBatchProcessorTest.class.getResource("/SqsFifoBatchEvent.json"), SQSEvent.class);
```
jeromevdl, line 40: why don't you use the https://github.com/aws/aws-lambda-java-libs/tree/main/aws-lambda-java-tests ?

jeromevdl (commented):
Tests are not easy to read, but probably because the code itself is... Few comments, Would be great to have another pair of eyes on it.

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
```diff
@@ -489,15 +493,41 @@ public static <R> List<R> batchProcessor(final SQSEvent event,
...
 
         for (SQSMessage message : event.getRecords()) {
-            try {
-                handlerReturn.add(handler.process(message));
-                batchContext.addSuccess(message);
-            } catch (Exception e) {
-                batchContext.addFailure(message, e);
-                LOG.error("Encountered issue processing message: {}", message.getMessageId(), e);
+            // If the batch has already failed, we mark each message past the failure point as failed
+            if (failWholeBatch) {
```
jeromevdl, line 505: Would it be possible to avoid looping over the remaining messages? Once it is failed, we add the other messages in the failures in one step after the loop (I did it in the [RFC](https://github.com/aws-powertools/powertools-lambda-java/blob/rfc-batch-processor/powertools-batch/src/main/java/software/amazon/lambda/powertools/batch2/SQSBatchProcessor.java#L75-L103))

powertools-sqs/src/test/java/software/amazon/lambda/powertools/sqs/SqsUtilsFifoBatchProcessorTest.java
scottgerring: Done! I load it once in the constructor (i've only got one event to use for now).

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
scottgerring: I'm not so sure about this - I find that variant quite hard to grok with the tracking of the offset into the list and the c-style cursor movement. I think we should err on the side of clarity as we're dealing with small lists. I could imagine another alternative could be:

1. Push everything into a queue
2. Pop one by one until you fail
3. Once you fail `break;` then push the rest into the failure list

WDYT @jeromevdl ?

jeromevdl: I kinda like what I did as an optimization... do whatever you prefer but please don't loop over failed message. (I find strange to push message into a queue, while they are already messages from a queue ;) )

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
```diff
@@ -490,19 +496,42 @@ public static <R> List<R> batchProcessor(final SQSEvent event,
...
-        for (SQSMessage message : event.getRecords()) {
+        Queue<SQSMessage> messagesToProcess = new LinkedList<>(event.getRecords());
+        while (!messagesToProcess.isEmpty()) {
+            SQSMessage message = messagesToProcess.remove();
+            // If the batch hasn't failed, try process the message
             try {
                 handlerReturn.add(handler.process(message));
                 batchContext.addSuccess(message);
-            } catch (Exception e) {
+            } catch(Exception e){
```
jeromevdl, line 506:
```suggestion
            } catch (Exception e) {
```

powertools-sqs/src/test/java/software/amazon/lambda/powertools/sqs/SqsUtilsFifoBatchProcessorTest.java
```diff
@@ -0,0 +1,124 @@
...
+    private MockitoSession session;
+
+    @Mock
+    private SqsClient sqsClient;
+
+    @Captor
+    private ArgumentCaptor<DeleteMessageBatchRequest> deleteMessageBatchCaptor;
+
+    public SqsUtilsFifoBatchProcessorTest() throws IOException {
+        //sqsBatchEvent  =  MAPPER.readValue(SqsUtilsFifoBatchProcessorTest.class.getResource("/SqsFifoBatchEvent.json"), SQSEvent.class);
```
jeromevdl, line 41: Remove comment

powertools-sqs/src/test/java/software/amazon/lambda/powertools/sqs/SqsUtilsFifoBatchProcessorTest.java
```diff
@@ -0,0 +1,124 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.MockitoAnnotations.openMocks;
+import static software.amazon.lambda.powertools.sqs.SqsUtils.batchProcessor;
+import static software.amazon.lambda.powertools.sqs.SqsUtils.overrideSqsClient;
+
+public class SqsUtilsFifoBatchProcessorTest {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
jeromevdl, line 30: Remove mapper

jeromevdl (commented):
Almost there

powertools-sqs/src/test/java/software/amazon/lambda/powertools/sqs/SqsUtilsFifoBatchProcessorTest.java
scottgerring: It would be nice if we had a linter that flagged PRs that introduce new unused variables

powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java
scottgerring: This should be the last detail @jeromevdl. I've tried to make the behaviour explicit, if verbose. I agree that shifting a pointer around will be more efficient. The maximum batch size of a regular queue is 10k or 10 for a FIFO queue. Maybe we should actually do it your way.

scottgerring: I read more and agree with you. 10k messages per batch maximum I'm happy to shift focus to efficiency. 

Here's the loop:

https://github.com/aws-powertools/powertools-lambda-java/blob/685177605bc9d5a96f4a1ea96102679c06e9475b/powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java#L498-L529

And here's capturing the failures:

https://github.com/aws-powertools/powertools-lambda-java/blob/685177605bc9d5a96f4a1ea96102679c06e9475b/powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/SqsUtils.java#L533-L541

@jeromevdl I think this should be it! Please have a bit of a dive in here, but I think the test coverage is good (and I added an extra one for this).

jeromevdl (commented):
just remove the break and we're good

jeromevdl (approved):
LGTM !!

Issue #1102, opened 2023-03-20
Maintenance: Report on the jar file size in PRs


jeromevdl:
### Summary

Have an action in GitHub workflow (PR, push to master) to calculate the size of the jar files and report it in the PR with the trend (increase/decrease).

### Why is this needed?

One of our tenet is "Keep it lean", to avoid adding too many dependencies and having powertools too heavy, causing slower cold starts. Having the information in each PR of the size and the trend would help making the good choices.

### Which area does this relate to?

Automation, Governance

### Solution

Could not find any Java-related GitHub actions (a lot on Js/Ts), but maybe something more generic exists.

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-java/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

jeromevdl:
done with #1196

PR #1196, opened 2023-06-15
chore: Reporting size of the jars in GitHub comments

jeromevdl:
**Issue #, if available:** #1102 

## Description of changes:

Display a report in GitHub comments about jars size. (Ideally would be great to have the history and see if it's growing)

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,14 +77,15 @@ jobs:
         id: artifacts-size-report
         if: ${{ matrix.java == '11.0.x' }} # do it once
         run: |
-          echo '## Artifacts Size Report' > report.md
-          echo '| Module | Version | Size |' >> report.md
+          echo '## :floppy_disk: Artifacts Size Report' > report.md
+          echo '| Module | Version | Size (KB) |' >> report.md
           echo '| --- | --- | --- |' >> report.md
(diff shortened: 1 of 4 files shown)


jeromevdl:
Help: How can we compare with previous versions of each artifact? (ideally from maven central or any other way)

scottgerring:
Hey @jeromevdl can we help here ?!

jeromevdl:
@scottgerring Yes, for the moment I report the size of the built artifacts. Ideally I'd like to compare with the current release to see if it grows or not, but not sure how to achieve this...

jeromevdl:
My idea for the comparison with previous versions:
- During release, generate a json file containing the artifacts size that we push to S3 (public access allowed on the S3 file).
- Then when there is a PR, we retrieve the json file to be able to compare.

We'll do that when refactoring the release process... We can already merge the report itself.

Ready for review!

scottgerring:
Can we do this with [sonar custom metrics](https://docs.sonarsource.com/sonarqube/8.9/instance-administration/custom-measures/)?

jeromevdl:
Will push that initial version without comparison with previous versions.
Actually it would be great to have a report of the new dependencies added... or a dependency tree or something to see if add weight to the final asset the user will deploy.

PR #1232, opened 2023-06-23
chore: secure github actions using hash instead of versions

jeromevdl:
**Issue #, if available:**

## Description of changes:

Replace all explicit versions with hashes to pin to a specific version.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


scottgerring (commented):
I think this is an improvement on how it is today and we should merge it.

I also think in a subsequent task it is worth digging into the smaller 3rd party actions - e.g. `ahmadnassri/action-workflow-run-wait`, `release-drafter/release-drafter`, and `jacobtomlinson/gha-find-replace` and checking if we can't get away with some github 1st party action, _or_, a powertools shared action, _or_ failing the previous options reviewing the code of the thing.

scottgerring (approved):

scottgerring:
@jeromevdl I think we should merge this - wdyt?

jeromevdl:
agree!

jeromevdl:
except that build fails without clear reason why...

jeromevdl:
> except that build fails without clear reason why...

It was because of the docs.yaml deleted by @kozub, the merge didn't work well... I've deleted it.

sthulb (approved):
This can be merged assuming the code at each of the hashes has been reviewed to be safe.

Issue #1224, opened 2023-06-21
Maintenance: Decouple samples from the parent pom


jeromevdl:
### Summary

Currently, samples (`powertools-examples`) are inheriting stuff from the parent pom (eg. aspectJ config and others) and are not really usable as is by developers. We should make them independent/freestanding so that anyone can take it without looking at our parent configuration.

### Why is this needed?

Current samples are not really reusable as is, developers would need to figure out what to take from our parent pom to put in their code.

### Which area does this relate to?

Other

### Solution

remove the following from https://github.com/aws-powertools/powertools-lambda-java/blob/main/examples/pom.xml: 

```xml
<parent>
    <artifactId>powertools-parent</artifactId>
    <groupId>software.amazon.lambda</groupId>
    <version>1.15.0</version>
</parent>
```

And make sure samples are still working, especially make sure to take the aspectj-maven-plugin configuration.

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda (Java) languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

scottgerring:
I think this makes sense. However we should make sure that the build still breaks if a change in the powertools modules breaks the examples. If we just remove the parent from the examples pom, I don't think this will be the case. Happy to be wrong on this one - just want to make sure that we don't end up in a situation where we can accidentally break the examples and not notice.

jeromevdl:
We can certainly keep samples as modules so that we build them with the rest. But we remove the `<parent>` flag to avoid inheriting stuff. And we put them last in the list of modules and change `package` with `install` to keep this "dependency" and make sure samples fail if there's an issue in powertools.

scottgerring:
> We can certainly keep samples as modules so that we build them with the rest. But we remove the `<parent>` flag to avoid inheriting stuff. And we put them last

Great! I didn't know you could have modules without the parent in the other direction. Exactly what we want. I think we should also do the same the next layer down - `parent -> samples parent -> [each example]` . That way you really can just grab a sample by itself and use it, rather than having to grab the whole `samples` tree.

> I also think we should adopt the x.y.z-SNAPSHOT notation while in development.

+1

[...]

PR #1240, opened 2023-06-25
[fix] Decouple samples from the parent pom

scottgerring:
**Issue #, if available:**
#1224 

## Description of changes:
The examples are two layers of maven module beneath the parent pom `parent --> examples --> [parameters / core / sqs / etc example projects]`. At each layer a dependency upwards on the immediate parent pom is included - the example projects themselves have dependencies on the examples common parent, and the examples common parent has a dependency on the overall parent. 
This change keeps the module composition, but removes the parents. This means that all the dependencies have to be specified in each example project, but the advantage is that the example projects can be taken and used as-is - they are completely self contained.

By keeping the module composition going down, we also ensure that the build continues to work as you would expect - changing the interface of the parameters project for instance will break the parameters example and the build, giving the developer immediate feedback. 

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


scottgerring:
@jeromevdl is it obvious to you [why this build is failing](https://github.com/aws-powertools/powertools-lambda-java/actions/runs/5369672123/jobs/9741278967?pr=1240) off the top of your head? If not I will dive deep.

Parameters example fails with:

```
Error:  Failed to execute goal dev.aspectj:aspectj-maven-plugin:1.13.1:compile (default) on project powertools-examples-core: Execution default of goal dev.aspectj:aspectj-maven-plugin:1.13.1:compile failed: An API incompatibility was encountered while executing dev.aspectj:aspectj-maven-plugin:1.13.1:compile: java.lang.UnsupportedClassVersionError: org/eclipse/core/runtime/OperationCanceledException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```
Which is the difference between Java 11 and Java 8. As far as I can tell from looking at the build, the whole thing should compile with Java 8, and when I build it locally with Corretto 1.8 it all works. Am I missing something obvious to you?

jeromevdl:
> Parameters example fails with:
> 
> ```
> Error:  Failed to execute goal dev.aspectj:aspectj-maven-plugin:1.13.1:compile (default) on project powertools-examples-core: Execution default of goal dev.aspectj:aspectj-maven-plugin:1.13.1:compile failed: An API incompatibility was encountered while executing dev.aspectj:aspectj-maven-plugin:1.13.1:compile: java.lang.UnsupportedClassVersionError: org/eclipse/core/runtime/OperationCanceledException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
> ```
> 
> Which is the difference between Java 11 and Java 8. As far as I can tell from looking at the build, the whole thing should compile with Java 8, and when I build it locally with Corretto 1.8 it all works. Am I missing something obvious to you?

This is exactly why I wanted to get samples autonomous... So that we can see this kind of things. This is the same error I had in end-to-end tests. You need to enforce the version 1.9.7 of aspecttools or aspectjrt, look at the e2e pom.

examples/pom.xml
```diff
@@ -4,15 +4,11 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
+    <groupId>software.amazon.lambda</groupId>
     <artifactId>powertools-examples</artifactId>
+    <version>1.15.0</version>
     <packaging>pom</packaging>
 
-    <parent>
```
scottgerring, line 10: Drop examples dependency on parent pom

examples/pom.xml
```diff
@@ -32,32 +28,4 @@
         <maven.deploy.skip>true</maven.deploy.skip>
     </properties>
 
-    <build>
```
scottgerring, line 35: We can drop this, because the coverage plugin is no longer configured by the parent pom in the examples, as they no longer extend it.

examples/powertools-examples-core/pom.xml
```diff
@@ -118,4 +122,72 @@
         </plugin>
       </plugins>
     </build>
+    <profiles>
```
scottgerring, line 125: Add a profile to override the aspectj setup for Java 1.8. All other java versions build happily with the default configuration.

examples/powertools-examples-core/pom.xml
```diff
@@ -118,4 +122,72 @@
...
+                        <plugin>
+                            <groupId>dev.aspectj</groupId>
+                            <artifactId>aspectj-maven-plugin</artifactId>
+                            <version>${aspectj.plugin.version}</version>
+                            <configuration>
+                                <source>${maven.compiler.source}</source>
+                                <target>${maven.compiler.target}</target>
+                                <complianceLevel>${maven.compiler.target}</complianceLevel>
+                                <aspectLibraries>
+                                    <aspectLibrary>
```
scottgerring, line 158: In each particular example, the set of `aspectLibraries` specified must be reflected in both the main pom and the profile override

scottgerring:
@jeromevdl have I missed anything?

*  Readme updated
* Projects build and weave independently for all java versions
* Breaking interfaces in powertools projects immediately fails build for examples 

You had a comment on the ticket that I've not done anything with that I might've missed the significance of: 
> and change package with install to keep this "dependency"

examples/pom.xml
jeromevdl: 👍

examples/powertools-examples-core/pom.xml
```diff
@@ -118,4 +122,72 @@
...
+                                        <artifactId>powertools-tracing</artifactId>
+                                    </aspectLibrary>
+                                    <aspectLibrary>
+                                        <groupId>software.amazon.lambda</groupId>
+                                        <artifactId>powertools-logging</artifactId>
+                                    </aspectLibrary>
+                                    <aspectLibrary>
+                                        <groupId>software.amazon.lambda</groupId>
+                                        <artifactId>powertools-metrics</artifactId>
+                                    </aspectLibrary>                                </aspectLibraries>
```
jeromevdl, line 169:
```suggestion
                                    </aspectLibrary>
                                </aspectLibraries>
```

examples/powertools-examples-core/pom.xml
```diff
@@ -118,4 +122,72 @@
         </plugin>
       </plugins>
     </build>
+    <profiles>
+        <!-- Use a profile to enforce AspectJ version 1.9.7 if we are Java 1.8, otherwise we'll get class
+             version mismatch issues. All subsequent Java releases build with the default AspectJ configuration
+              on the project. -->
```
jeromevdl, line 128:
```suggestion
              on the project. If you use Java8, simply add the aspectjtools 1.9.7 to your dependencies and the plugin dependencies. This profiles is not needed if you target only one JDK -->
```

examples/powertools-examples-idempotency/pom.xml
```diff
@@ -168,6 +171,75 @@
         </plugin>
       </plugins>
     </build>
+    <profiles>
+        <!-- Use a profile to enforce AspectJ version 1.9.7 if we are Java 1.8, otherwise we'll get class
+             version mismatch issues. All subsequent Java releases build with the default AspectJ configuration
+              on the project. -->
```
jeromevdl, line 177:
```suggestion
              on the project. If you use Java8, simply add the aspectjtools 1.9.7 to your dependencies and the plugin dependencies. This profiles is not needed if you target only one JDK. -->
```

examples/powertools-examples-parameters/pom.xml
```diff
@@ -87,4 +88,62 @@
         </plugin>
       </plugins>
     </build>
+    <profiles>
+        <!-- Use a profile to enforce AspectJ version 1.9.7 if we are Java 1.8, otherwise we'll get class
+             version mismatch issues. All subsequent Java releases build with the default AspectJ configuration
+              on the project. -->
```
jeromevdl, line 94:
```suggestion
              on the project. If you use Java8, simply add the aspectjtools 1.9.7 to your dependencies and the plugin dependencies. This profiles is not needed if you target only one JDK. -->
```

examples/powertools-examples-serialization/pom.xml
```diff
@@ -64,4 +65,63 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <!-- Use a profile to enforce AspectJ version 1.9.7 if we are Java 1.8, otherwise we'll get class
+             version mismatch issues. All subsequent Java releases build with the default AspectJ configuration
+              on the project. -->
```
jeromevdl, line 72:
```suggestion
              on the project. If you use Java8, simply add the aspectjtools 1.9.7 to your dependencies and the plugin dependencies. This profiles is not needed if you target only one JDK. -->
```

examples/powertools-examples-sqs/pom.xml
```diff
@@ -119,4 +121,69 @@
         </plugin>
       </plugins>
     </build>
+    <profiles>
+        <!-- Use a profile to enforce AspectJ version 1.9.7 if we are Java 1.8, otherwise we'll get class
+             version mismatch issues. All subsequent Java releases build with the default AspectJ configuration
+              on the project. -->
```
jeromevdl, line 127:
```suggestion
              on the project. If you use Java8, simply add the aspectjtools 1.9.7 to your dependencies and the plugin dependencies. This profiles is not needed if you target only one JDK. -->
```

examples/powertools-examples-validation/pom.xml
```diff
@@ -82,4 +83,65 @@
         </plugin>
       </plugins>
     </build>
+    <profiles>
+        <!-- Use a profile to enforce AspectJ version 1.9.7 if we are Java 1.8, otherwise we'll get class
+             version mismatch issues. All subsequent Java releases build with the default AspectJ configuration
+              on the project. -->
```
jeromevdl, line 89:
```suggestion
              on the project. If you use Java8, simply add the aspectjtools 1.9.7 to your dependencies and the plugin dependencies. This profiles is not needed if you target only one JDK. -->
```

jeromevdl (commented):
Just suggesting to add a comment on the fact that profiles are not really required by developers. You can probably write it with better english ;)

jeromevdl:
> You had a comment on the ticket that I've not done anything with that I might've missed the significance of:
> 
> > and change package with install to keep this "dependency"

@scottgerring, I was mentioning we should change the `mvn package` to `mvn install` so that examples will take the "installed" versions of powertools (locally), not the one in maven central. Thus we keep the dependency, and samples can fail if there is something wrong in the new version of Powertools.

scottgerring:
@jeromevdl cleaned up comments and changed github CI build to run `install`

jeromevdl:
You didn't like my comments ? maybe you can reformulate but I think we should say they don't really need the profiles, but just to take the dependency...

scottgerring:
> You didn't like my comments ? maybe you can reformulate but I think we should say they don't really need the profiles, but just to take the dependency...

my bad; I failed to push the [commit resolving this[(https://github.com/aws-powertools/powertools-lambda-java/pull/1240/commits/9629436585b1d40aa9939aee10263fe10b38d6fd) yesterday before resolving the threads 🤦 I expanded on the profile xmldoc in each example like this: 

```xml
  <!-- Use a profile to enforce AspectJ version 1.9.7 if we are Java 1.8 otherwise we'll get class
             version mismatch issues. All subsequent Java releases build with the default AspectJ configuration
              on the project.

              Note:
              - if you are running Java > 1.8, you can remove this profile altogether
              - If you are running on Java 1.8, you should apply the aspectJ version here to the project, and remove
                the profile.
              -->
```

jeromevdl (approved):
Good for me

Issue #1281, opened 2023-07-12
Idempotency in progress timeout


scottgerring:
Two expirations are stored on the DDB table:

* expiration - how long to cache the results of the operation. User-tunable  ( seconds since epoch )
* in_progress_expiration - expire records that timed out during lambda execution. Set by powertools to the remainingExecutionTime.  ( milliseconds since epoch)

When a request comes in, we try and conditionally upsert the DDB record - if the `in_progress_expiration` time is before `now()`, we can overwrite it, else we can't. The bug is here:

https://github.com/aws-powertools/powertools-lambda-java/blob/a19def3e4ea9e510c0ce4ad602abe58db0610d1f/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java#L154

https://github.com/aws-powertools/powertools-lambda-java/blob/a19def3e4ea9e510c0ce4ad602abe58db0610d1f/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java#L163

We set `now` once, but we use it in a comparison where the units should be seconds, and another where the units should be milliseconds. 

The upshot of this is that if a handler times out within an idempotent operation, we'll start seeing exceptions thrown for subsequent invocations. 

**What were you trying to accomplish?**

## Expected Behavior
Subsequent executions after a function timeout should update the DDB table and complete. 

## Possible Solution
1. Provide `nowMs` and `now` in the DDB conditional update
2. Increase test coverage

PR #1285, opened 2023-07-13
fix: idempotency timeout bug

scottgerring:
**Issue #, if available:**
#1281 

## Description of changes:
Fixes a bug in idempotency where the processing of a idempotent record can become stalled after an earlier Lambda timeout. 

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DataRecord.java
+++ b/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DataRecord.java
@@ -28,7 +28,7 @@ public class DataRecord {
     private final String status;
 
     /**
-     * This field is control how long the result of the idempotent
+     * This field is controlling how long the result of the idempotent
      * event is cached. It is stored in _seconds since epoch_.
      *
      * DynamoDB's TTL mechanism is used to remove the record once the
--- a/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DataRecord.java
+++ b/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DataRecord.java
@@ -26,9 +26,30 @@ import java.util.OptionalLong;
 public class DataRecord {
     private final String idempotencyKey;
     private final String status;
+
+    /**
+     * This field is control how long the result of the idempotent
+     * event is cached. It is stored in _seconds since epoch_.
+     *
+     * DynamoDB's TTL mechanism is used to remove the record once the
+     * expiry has been reached, and subsequent execution of the request
+     * will be permitted. The user must configure this on their table.
+     */
     private final long expiryTimestamp;
     private final String responseData;
     private final String payloadHash;
+
+    /**
+     * The in-progress field is set to the remaining lambda execution time
+     * when the record is created.
+     * This field is stored in _milliseconds since epoch_.
+     *
+     * This ensures that:
+     *
+     * 1/ other concurrently executing requests are blocked from starting
+     * 2/ if a lambda times out, subsequent requests will be allowed again, despite
+     *     the fact that the idempotency record is already in the table
+     */
     private final OptionalLong inProgressExpiryTimestamp;
 
     public DataRecord(String idempotencyKey, Status status, long expiryTimestamp, String responseData, String payloadHash) {
--- a/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java
+++ b/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java
@@ -152,6 +152,7 @@ public class DynamoDBPersistenceStore extends BasePersistenceStore implements Pe
 
             Map<String, AttributeValue> expressionAttributeValues = Stream.of(
                     new AbstractMap.SimpleEntry<>(":now", AttributeValue.builder().n(String.valueOf(now.getEpochSecond())).build()),
+                    new AbstractMap.SimpleEntry<>(":now_milliseconds", AttributeValue.builder().n(String.valueOf(now.getEpochSecond() * 1000)).build()),
                     new AbstractMap.SimpleEntry<>(":inprogress", AttributeValue.builder().s(INPROGRESS.toString()).build())
             ).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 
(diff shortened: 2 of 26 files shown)


powertools-idempotency/src/test/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStoreTest.java
```diff
@@ -95,7 +95,7 @@ public void putRecord_shouldCreateRecordInDynamoDB_IfLambdaWasInProgressAndTimed
         Map<String, AttributeValue> item = new HashMap<>(key);
         Instant now = Instant.now();
         long expiry = now.plus(30, ChronoUnit.SECONDS).getEpochSecond();
-        long progressExpiry = now.minus(30, ChronoUnit.SECONDS).getEpochSecond();
+        long progressExpiry = now.minus(30, ChronoUnit.SECONDS).toEpochMilli();
```
scottgerring, line 98: This test covers the "timed out" behaviour, but had to be fixed for the units

powertools-idempotency/src/test/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStoreTest.java
```diff
@@ -152,14 +152,14 @@ public void putRecord_shouldThrowIdempotencyItemAlreadyExistsException_IfRecordA
...
-    public void putRecord_shouldThrowIdempotencyItemAlreadyExistsException_IfRecordAlreadyExistAndProgressNotExpiredAfterLambdaTimedOut() {
+    public void putRecord_shouldBlockUpdate_IfRecordAlreadyExistAndProgressNotExpiredAfterLambdaTimedOut() {
         key = Collections.singletonMap("id", AttributeValue.builder().s("key").build());
 
         // GIVEN: Insert a fake item with same id
         Map<String, AttributeValue> item = new HashMap<>(key);
         Instant now = Instant.now();
         long expiry = now.plus(30, ChronoUnit.SECONDS).getEpochSecond(); // not expired
-        long progressExpiry = now.plus(30, ChronoUnit.SECONDS).getEpochSecond(); // not expired
+        long progressExpiry = now.plus(30, ChronoUnit.SECONDS).toEpochMilli(); // not expired
```
scottgerring, line 162: This test covers the case where the update should be blocked, but again needs to be fixed for units

powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java
```diff
@@ -126,6 +126,19 @@ public DataRecord getRecord(String idempotencyKey) throws IdempotencyItemNotFoun
         return itemToRecord(response.item());
     }
 
+    /**
+     * Store's the given idempotency record in the DDB store. If there
```
scottgerring, line 130: Trying to make it clearer what this method is doing - both the conditional updating of the existing record and the fact that it doesn't write the _complete_ DataRecord seems important.

powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DataRecord.java
```diff
@@ -26,9 +26,30 @@
...
+     * The in-progress field is set to the remaining lambda execution time
+     * when the record is created.
+     * This field is stored in _milliseconds since epoch_.
+     *
+     * This ensures that:
+     *
+     * 1/ other concurrently executing requests are blocked from starting
+     * 2/ if a lambda times out, subsequent requests will be allowed again, despite
+     *     the fact that the idempotency record is already in the table
+     */
```
scottgerring, line 52: I could also rename this field completely - `inProgressExpiryTimestampMs`. It's not public. WDYT?

scottgerring:
I've gone through all uses of the `DataRecord` constructor and `DataRecord::getExpiryTimestamp` to make sure there are no other unit mismatch surprises. I think this covers it!

powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DataRecord.java
```diff
@@ -26,9 +26,30 @@
 public class DataRecord {
     private final String idempotencyKey;
     private final String status;
+
+    /**
+     * This field is control how long the result of the idempotent
```
jeromevdl, line 31:
```suggestion
     * This field is controlling how long the result of the idempotent
```

powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java
```diff
@@ -152,6 +165,7 @@ public void putRecord(DataRecord record, Instant now) throws IdempotencyItemAlre
 
             Map<String, AttributeValue> expressionAttributeValues = Stream.of(
                     new AbstractMap.SimpleEntry<>(":now", AttributeValue.builder().n(String.valueOf(now.getEpochSecond())).build()),
+                    new AbstractMap.SimpleEntry<>(":now_milliseconds", AttributeValue.builder().n(String.valueOf(now.getEpochSecond() * 1000)).build()),
```
jeromevdl, line 168: why not using `toEpochMilli()` instead of doing math?

powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DataRecord.java
scottgerring: I've changed my mind on this. Because the setter/getter are public (should they be?) it doesn't make much sense to rename the field alone.

scottgerring: I've changed my mind on this. Because the setter/getter are public (should they be?) it doesn't make much sense to rename the field alone.

jeromevdl: let's keep them like this, it would be strange to specify the unit for one and not the other.

jeromevdl (approved):
good for me!

scottgerring:
Retested using powertools-examples-idempotency as well as the fixed tests in here. This is good to go.

Issue #1302, opened 2023-07-20
Initialization of SDK client in `DynamoDBPersistenceStore` leads to an error when used with SnapStart under certain conditions 


roamingthings:
Yesterday one of my Lambda Function that is deployed using SnapStart ran into a problem that has been caused by
code that has been derived from this repository.
```java
// AWS_LAMBDA_INITIALIZATION_TYPE has two values on-demand and snap-start
// when using snap-start mode, the env var creds provider isn't used and causes a fatal error if set
// fall back to the default provider chain if the mode is anything other than on-demand.
String initializationType = System.getenv().get(AWS_LAMBDA_INITIALIZATION_TYPE);
if (initializationType  != null && initializationType.equals(ON_DEMAND)) {
    ddbBuilder.credentialsProvider(EnvironmentVariableCredentialsProvider.create());
}
```
https://github.com/aws-powertools/powertools-lambda-java/blob/c185cb063a4612f704d092fb21b2493706baa193/powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java#L100

An SDK client that has been initialized like that may fail to retrieve the credentials after an invocation has been timed out and the runtime is reused by subsequent invocations.

Also the error has been visible in my code, the code in this library may also be affected.

## Expected Behavior

The client initialized in`DynamoDBPersistenceStore` will be able to retrieve its credentials after an previous invocation of the same runtime timed out.

## Current Behavior

`DynamoDBPersistenceStore` fails when a previous invocation timed out.

## Possible Solution

Always use the default credentials provider chain by removing the code shown above. This has also been the suggested solution provided by AWS support.

## Steps to Reproduce (for bugs)

I've created a proof-of-concept project to reproduce the issue:
https://github.com/roamingthings/sdk-client-snapstart-error-poc

What probably happens is the following:
* the function is initialized when publishing a new version.
  * `AWS_LAMBDA_INITIALIZATION_TYPE` is `snap-start`
* the SDK client is configured to use the default credentials provider chain
* the snapshot is generated
* the function gets invoked
* the snapshot is resumed
* the function times out for any reason (not related to this issue)
* the function is invoked again
* the runtime re-instantiates the handler
  *   * `AWS_LAMBDA_INITIALIZATION_TYPE` is `on-demand`
* the SDK client is also re-instantiated (it's not static in our case)
* the SDK client is configured to only use the environment variable credentials provider
[...]

PR #1303, opened 2023-07-20
fix: use default credentials provider for all provided SDK clients

roamingthings:
**Issue #, if available:** #1302

## Description of changes:

This pull request removes the conditional configuration of the credentials provider for all SDK clients that are created when no other SDK client has been provided by the application. As a consequence these SDK clients will always use the credentials provider chain that is used by SDK clients by default.

This change is necessary because in some situations a Lambda function that is published using SnapStart will be re-initialized by the runtime with its invocation type set to `on-demand`. As a consequence the SDK client would be configured to only use the `EnvironmentVariableCredentialsProvider`. The next time the client is used an error will be thrown since the credentials are not provided differently than without SnapStart.

For more details please see #1302 and http://github.com/roamingthings/sdk-client-snapstart-error-poc

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


scottgerring:
@msailes as you are closest to Snapstart and the original change it would be great if you could help us convince ourselves this is the best way to deal with this.

IMHO it would also be nice if we can add an E2E test for this - we have the framework for it now running in the repo, and @roamingthings has given us a reproducible [deployable test case ](https://github.com/roamingthings/sdk-client-snapstart-error-poc)

roamingthings:
I have to add that I have a support ticket open to understand if the behavior of the runtime (cold-start after timeout) is intentional.

scottgerring:
Hey @roamingthings to keep you posted - we're asking around internally to make sure that the the default provider chain is the best way to handle this. Concretely this means making sure that we don't take a big performance hit when we work through it, and that there isn't some other best practice we should be employing. We're also having a bit of a look to see if we can add to our E2E tests to catch these sort of issues in the future.

roamingthings:
@scottgerring Thank you for keeping me updated. At least support told me that this is the way to go. I'm curious what you will find out. Since there is a workaround for this issue I'm not in a hurry.

mriccia:
Thank you @roamingthings 
I took [your example](https://github.com/roamingthings/sdk-client-snapstart-error-poc) and I used it to run some Load Tests, both with the changes in this PR, and with version `1.16.1`. (Note: I made some modifications to run without SnapStart and removing the timeout)

I confirm that with these changes there is no noticeable performance difference, you can find a summary of my tests below:
### Version 1.16.1
```
Summary report @ 19:03:02(+0200) 2023-07-21
  Scenarios launched:  8250
  Scenarios completed: 4745
  Requests completed:  4745
  Mean response/sec: 36.58
  Response time (msec):
    min: 51
    max: 5678
    median: 81
    p95: 156.3
    p99: 5372.5
  Scenario counts:
    0: 8250 (100%)
  Codes:
    200: 4745
  Errors:
    ETIMEDOUT: 3505
 ```
 
 ### Version 1.17.0-SNAPSHOT
```
Summary report @ 18:57:09(+0200) 2023-07-21
  Scenarios launched:  8246
  Scenarios completed: 4449
  Requests completed:  4449
  Mean response/sec: 36.6
  Response time (msec):
    min: 54
    max: 5923
    median: 80
    p95: 157
    p99: 5393.1
  Scenario counts:
    0: 8246 (100%)
  Codes:
    200: 4449
  Errors:
    ETIMEDOUT: 3797
```

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaConstants.java
mriccia, line 1: Please let's keep the constants in this files. They are `public`, and users may be already using them

roamingthings:
For reference this describes the behavior of the runtime when an error like a timeout occurs during the invoke phase. This matches the observed behavior. 

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html#runtimes-lifecycle-restore

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaConstants.java
roamingthings: I agree.

Would it make sense to mark them as deprecated?

scottgerring: > I agree.
> 
> Would it make sense to mark them as deprecated?

That'd be great! We're starting to look at a v2 of the library and `@Deprecated` will make sure we don't miss this.

scottgerring:
@mriccia thanks for the data! Looks like we can make this change without fearing a performance regression.

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaConstants.java
roamingthings: I will add an annotation. Not sure if it would make sense to add any comment about the issue in the code.

scottgerring (approved):
LGTM!

mriccia (approved):
LGTM, thank you for spotting this issue and for writing this PR, @roamingthings!

Issue #1274, opened 2023-07-09
Maintenance: append user agent to AWS SDK clients


jeromevdl:
### Summary

The AWS SDK v2 allows to add user agent headers to the requests made using its clients.

### Why is this needed?

So that AWS can attribute requests made to its API when using Powertools.

### Which area does this relate to?

Governance

### Solution

Expose a shared function (in core module) that takes a client and adds Powertools-related user agents to the request. This shared function can then be used by other Powertools utilities that make requests to AWS APIs. Check [SDK documentation](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/client/config/SdkAdvancedClientOption.html) and [typescript implementation](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/commons/src/userAgentMiddleware.ts) (and [usage](https://github.com/aws-powertools/powertools-lambda-typescript/blob/2eb7ab3cb84e9f57378ca8f3c8d27e7912c18a3d/packages/idempotency/src/persistence/DynamoDBPersistenceLayer.ts#L92)). 
 
Impacted modules:
- tracing (X-ray)
- parameters (SSM Parameters, Secrets Manager, DynamoDB, AppConfig)
- idempotency (DynamoDB)
- sqs (SQS)

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda (Java) languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

eldimi:
I will have a look into this

scottgerring:
Thanks @eldimi - I'll assign it to you then so we can keep an overview of what's going on :D

Issue #1092, opened 2023-03-16
Move SDK http client to CRT


jeromevdl:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
As the CRT http client is now available (https://aws.amazon.com/blogs/developer/announcing-availability-of-the-aws-crt-http-client-in-the-aws-sdk-for-java-2-x/), we could use it in the modules we use the SDK (Parameters, Idempotency, SQS, CloudFormation) to reduce the cold start and latency.

Due to the scale, we should do this on one module first for review, and then go from there. Idempotency would be a good candidate. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

codesometech:
@jeromevdl @msailes I can pick this up

jreijn:
I would like to pick this up if possible?

jeromevdl:
Hi @jreijn , we're not yet ready for this.

scottgerring:
We've setup the v2 branch so that we're ready to start taking v2 contributions ( #1337 ) and we can start working on this. Following on from What Eleni linked, we'll want to set this up so that common initialization of the clients can be configured in one place in the `core` module. That is - we should see if we can have one place that does stuff like setting `User-Agent`, region configuration, etc.), and so on, rather than duplicating this everywhere. This would have to work without coupling `powertools-core` onto all of the other SDKs of course.

Speaking with @jeromevdl , we think this is going to be a big, cross-cutting change, because CRT imposes async on everything. We'd love to see a single module setup with it, so we have a point of discussion, rather than big-banging the whole thing.

@jreijn are you still interested in helping here?

jreijn:
@scottgerring yes, I would still like to work on this. Not sure if there is any time planning for the release of v2 as I will be out for the next 2.5 weeks. Otherwise I will continue with my work on this after my vacation.

I agree with taking a single module setup, so we can discuss while moving forward.

scottgerring:
Hey @jreijn , we'd like to get a bit more definitive with our release planning, but I don't think that being off for a few weeks  now makes a difference! Can you reach out when you are back? Suggest we have a quick kick-off chat. Will start having a think about this in the meantime. 

Enjoy the break!

msailes:
Lets watch this issue in the AWS SDK for Java v2 https://github.com/aws/aws-sdk-java-v2/issues/3343

scottgerring:
@msailes do you have any insight into the priority of this?

[...]

PR #1306, opened 2023-07-21
chore: Add powertools specific user-agent-suffix to the AWS SDK v2 clients

eldimi:
**Issue #:**
https://github.com/aws-powertools/powertools-lambda-java/issues/1274

## Description of changes:

**Checklist**

 

* [x] Meet tenets criteria](https://docs.powertools.aws.dev/lambda-java/#tenets)
* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
```diff
@@ -0,0 +1,90 @@
...
+import com.sun.org.slf4j.internal.LoggerFactory;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.net.URL;
+import java.util.Properties;
+
+import static software.amazon.lambda.powertools.core.internal.SystemWrapper.getenv;
+
+public class UserAgentConfigurer {
```
scottgerring, line 13:
```suggestion
public class UserAgentConfigurator {
```

powertools-core/src/main/resources/version.properties
```diff
@@ -0,0 +1 @@
+powertools.version=${project.version}
```
scottgerring, line 1: Will this be filled by the build as it exists currently or do we need to do something special?

powertools-core/src/test/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurerTest.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.mockito.Mockito.mockStatic;
+import static software.amazon.lambda.powertools.core.internal.SystemWrapper.getenv;
+import static software.amazon.lambda.powertools.core.internal.UserAgentConfigurer.VERSION_KEY;
+import static software.amazon.lambda.powertools.core.internal.UserAgentConfigurer.getVersionFromProperties;
+import static software.amazon.lambda.powertools.core.internal.UserAgentConfigurer.VERSION_PROPERTIES_FILENAME;
+import static software.amazon.lambda.powertools.core.internal.UserAgentConfigurer.AWS_EXECUTION_ENV;
+
+
+
+public class UserAgentConfigurerTest {
```
scottgerring, line 19:
```suggestion
public class UserAgentConfiguratorTest {
```

powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java
```diff
@@ -90,13 +94,14 @@ private DynamoDBPersistenceStore(String tableName,
...
                         .httpClient(UrlConnectionHttpClient.builder().build())
+                        .overrideConfiguration(ClientOverrideConfiguration.builder().putAdvancedOption(SdkAdvancedClientOption.USER_AGENT_SUFFIX, UserAgentConfigurer.getUserAgent(IDEMPOTENCY)).build())
                         .region(Region.of(System.getenv(AWS_REGION_ENV)));
 
                 // AWS_LAMBDA_INITIALIZATION_TYPE has two values on-demand and snap-start
                 // when using snap-start mode, the env var creds provider isn't used and causes a fatal error if set
                 // fall back to the default provider chain if the mode is anything other than on-demand.
                 String initializationType = System.getenv().get(AWS_LAMBDA_INITIALIZATION_TYPE);
-                if (initializationType  != null && initializationType.equals(ON_DEMAND)) {
+                if (initializationType != null && initializationType.equals(ON_DEMAND)) {
```
scottgerring, line 104: I see this is just a syntax change, but we are about to change the credential chain as part of #1303 and we'll need to keep on the two merges.

pom.xml
```diff
@@ -519,6 +519,7 @@
                                 @{argLine}
                                 --add-opens java.base/java.util=ALL-UNNAMED
                                 --add-opens java.base/java.lang=ALL-UNNAMED
+                                --add-exports=java.xml.crypto/com.sun.org.slf4j.internal=ALL-UNNAMED
```
scottgerring, line 522: Not sure what to make of this?

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
scottgerring: And a bit of javadoc would be nice - this is here so that we can add extra info to the user agent in the AWS SDK sessions.

The python impl and docs are here --> 
https://github.com/aws-powertools/powertools-lambda-python/blob/021076efab85cb0afa867a0ffa68060917733f99/aws_lambda_powertools/shared/user_agent.py#L75

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final String NO_OP = "no-op";
+    private static String PT_VERSION = getProjectVersion();
+    private static String USER_AGENT_PATTERN = "PT/" + PT_FEATURE_VARIABLE + "/" + PT_VERSION + " PTEnv/" + PT_EXEC_ENV_VARIABLE;
+
+    /**
+     * Retrieves the project version from the version.properties file
+     *
+     * @return the project version
+     */
+    static String getProjectVersion() {
```
scottgerring, line 31: Can this be private?

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Retrieves the project version from a properties file.
+     * The file should be in the resources folder.
+     * The version is retrieved from the property with the given key.
+     *
+     * @param propertyFileName the name of the properties file
+     * @param versionKey       the key of the property that contains the version
+     * @return the version of the project as configured in the given properties file
+     */
+    static String getVersionFromProperties(String propertyFileName, String versionKey) {
```
scottgerring, line 45: Can this be private?

powertools-core/src/main/resources/version.properties
am29d, line 1: Does the value come from `pom.xml` during runtime?

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
```diff
@@ -0,0 +1,90 @@
...
+     * If it AWS_EXECUTION_ENV is not set, PT_EXEC_ENV defaults to "NA"
+     *
+     * @param ptFeature a custom feature to be added to the user agent string (e.g. idempotency).
+     *                  If null or empty, the default PT_FEATURE is used.
+     *                  The default PT_FEATURE is "no-op".
+     * @return the user agent string
+     */
+    public static String getUserAgent(String ptFeature) {
+
+        String awsExecutionEnv = getenv(AWS_EXECUTION_ENV);
```
scottgerring, line 78: Can this be fetched once statically too? - environment variables shouldn't change

powertools-core/src/test/resources/unreadable.properties
```diff
@@ -0,0 +1 @@
+powertools.version=${project.version}
```
am29d, line 1: Shouldn't this value be swapped with the `test.proprerties` empty value?

scottgerring (commented):
Looking good! Some comments inline.

am29d (commented):
Looks great! I only have two questions:

1. Can you also instrument clients that customer pass to the utility? Atm it only happens in `if (client == null)` blocks. 

2. Is it feasible to extract `overrideConfiguration` in a dedicated place? Atm every client runs exactly the same line. Might be easier to have it in one place and apply changes in the future in one file, instead of going through every constructor.

powertools-core/src/main/resources/version.properties
eldimi: [this](https://github.com/eldimi/aws-lambda-powertools-java/blob/chore/user_agent_sdk_clients/powertools-core/pom.xml#L101) should do the trick. It is already tested in the unit tests. I'll import the library in a project to test again and make sure I didn't miss anything

powertools-core/src/test/resources/unreadable.properties
eldimi: The content is not really needed in this case, I can remove it if it is confusing. The file is [removed read access](https://github.com/eldimi/aws-lambda-powertools-java/blob/chore/user_agent_sdk_clients/powertools-core/src/test/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurerTest.java#L52) in a unit test to ensure version is set to "NA" in that case

powertools-core/src/main/resources/version.properties
eldimi: see my comment [here](https://github.com/aws-powertools/powertools-lambda-java/pull/1306#discussion_r1270759687)

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
eldimi: this for test purposes left to package access. It can get moved to a dedicated package though..

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
eldimi: same as [here](https://github.com/aws-powertools/powertools-lambda-java/pull/1306#discussion_r1270771817)

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
eldimi: Yes, it can, but I moved it here so that I can add a unit test for it. When it is a static variable it gets initializes before being mocked in the [test](https://github.com/eldimi/aws-lambda-powertools-java/blob/chore/user_agent_sdk_clients/powertools-core/src/test/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurerTest.java#L92)

powertools-core/src/test/resources/unreadable.properties
am29d: Ah interesting. It's a bit odd, because it looks that a valid input, but the name of the file does not match semantically, but I see now why it's called `unreadable`.

jeromevdl:
> Looks great! I only have two questions:
> 
> 1. Can you also instrument clients that customer pass to the utility? Atm it only happens in `if (client == null)` blocks.
> 2. Is it feasible to extract `overrideConfiguration` in a dedicated place? Atm every client runs exactly the same line. Might be easier to have it in one place and apply changes in the future in one file, instead of going through every constructor.

Overall looks good but I agree with Alex, we should add the user agent even on clients they pass to our modules. And +1 on the second point: can we have a method in the core module where we can pass any client, we override the configuration and return the client ?

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
scottgerring: Nah, I think that's reasonable as it is!

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
scottgerring: I dug into this a bit more. I thought this would be a straight pass-through to the underlying syscall and then it'd be nice if we didn't do that all the time. But - it turns out that in Java it's loaded once at startup and cached in a HashMap.

powertools-core/src/main/resources/version.properties
scottgerring: Probably worth publishing locally and then unzipping the jar and making sure that it contains `1.17.0-SNAPSHOT` in there

pom.xml
eldimi: yes, I had a wrong import, I noticed the failed build and [fixed]([b1809cc](https://github.com/aws-powertools/powertools-lambda-java/pull/1306/commits/b1809ccbd7d9dea551fcbbc9226abe85a85ba585))

powertools-idempotency/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/DynamoDBPersistenceStore.java
scottgerring: FYI - #1303 has been merged to main; if you merge down now you'll get this change

powertools-core/src/test/resources/unreadable.properties
eldimi: I added [comments](https://github.com/eldimi/aws-lambda-powertools-java/blob/chore/user_agent_sdk_clients/powertools-core/src/test/resources/unreadable.properties) also for that one, since it was confusing

powertools-core/src/main/java/software/amazon/lambda/powertools/core/internal/UserAgentConfigurer.java
eldimi: Done!

eldimi:
> > Looks great! I only have two questions:
> > 
> > 1. Can you also instrument clients that customer pass to the utility? Atm it only happens in `if (client == null)` blocks.

This can happen if we implement the [ExecutionInterceptor](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.html) and modify the request headers, since we can't modify the sdk client as soon as it is built. If we go this way, maybe it makes sense to do it in every case (whether the client already exists or not) so that we have a common implementation. What do you think?

> > 2. Is it feasible to extract `overrideConfiguration` in a dedicated place? Atm every client runs exactly the same line. Might be easier to have it in one place and apply changes in the future in one file, instead of going through every constructor.

I thought about doing that initially. But then, I thought that it might be better if we do a refactoring(in a seperate PR) and have a common way of building clients. This is because, duplicate code is also the issue when setting the region and the HttpClient. They are set is several places, the same way. So why extract only the overrideConfguration part? And if we modify something else in the future (either in overrideConfiguration or somewhere else), where should it be placed?
Thinking out loud here :)

> 
> Overall looks good but I agree with Alex, we should add the user agent even on clients they pass to our modules. And +1 on the second point: can we have a method in the core module where we can pass any client, we override the configuration and return the client ?
> 
> And the build is failing, not sure about the logger you use... (sun ? see [Scott's comment](https://github.com/aws-powertools/powertools-lambda-java/pull/1306/files#r1270734224) about that)

Yes, fixed that already, had accidentaly a wrong import and I didn't notice at all that was the case

@am29d @jeromevdl See my comments inline

powertools-core/src/main/resources-filtered/version.properties
```diff
@@ -0,0 +1,9 @@
+# The filtered properties can have variables that are filled in y system properties or project properties.
```
jeromevdl:
```suggestion
# The filtered properties can have variables that are filled in system properties or project properties.
```

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDbProvider.java
```diff
@@ -6,13 +6,8 @@
...
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
-import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
-import software.amazon.awssdk.services.dynamodb.model.GetItemRequest;
-import software.amazon.awssdk.services.dynamodb.model.GetItemResponse;
-import software.amazon.awssdk.services.dynamodb.model.QueryRequest;
-import software.amazon.awssdk.services.dynamodb.model.QueryResponse;
-import software.amazon.lambda.powertools.core.internal.LambdaConstants;
-import software.amazon.lambda.powertools.core.internal.UserAgentConfigurer;
+import software.amazon.awssdk.services.dynamodb.model.*;
```
jeromevdl, line 9: no "*", please expand

jeromevdl (commented):
almost OK

jeromevdl:
@eldimi I agree we should refactor this and have the common sdk client stuff in the core module ("common" in v2). Would you like to do this? (in another PR)

eldimi:
> @eldimi I agree we should refactor this and have the common sdk client stuff in the core module ("common" in v2). Would you like to do this? (in another PR)

yes, I can do that

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/DynamoDbProvider.java
eldimi: ohh! I did again! I had configured my IDE to not do that, but I might have missed something. Will fix that.

eldimi:
> almost OK

I saw some code smells reported by sonarqube - related to the current PR changes. I will fix them too in a next commit

eldimi:
> 1. instrument clients that customer



> almost OK

What abou instrumenting the clients that customers provide? Any comments on that related to my reply [here](https://github.com/aws-powertools/powertools-lambda-java/pull/1306#issuecomment-1651988426)?

jeromevdl:
> > 1. instrument clients that customer
> 
> > almost OK
> 
> What abou instrumenting the clients that customers provide? Any comments on that related to my reply [here](https://github.com/aws-powertools/powertools-lambda-java/pull/1306#issuecomment-1651988426)?

I agree, should be part of the refactor I mentioned. We should first merge this PR and then you move all the common stuff, plus the interceptor somewhere.

powertools-core/src/main/resources/version.properties
scottgerring: > Probably worth publishing locally and then unzipping the jar and making sure that it contains `1.17.0-SNAPSHOT` in there

I scratched my own itch and checked this from your fork - looks good - here's version.properties in the core jar:

```
# The filtered properties can have variables that are filled in by system properties or project properties.
# See https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html
#
# The values are replaced before copying the resources to the main output directory. Therefore, as soon as the build phase is completed,
# the values should have been replaced if the properties are available and if 'filtering' is set to true in the pom.xml
#
#
# The 1.17.0-SNAPSHOT is retrieved from the respective pom.xml property
powertools.version=1.17.0-SNAPSHOT
```

eldimi:
Is there anything missing from my side here?

eldimi:
> @eldimi, I'll merge that one. Can you write an RFC for the instrumentation and generic client stuff you proposed? So we can elaborate on it...

Looking a bit more into that..The [ExecutionInterceptor](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.html) is only configured in the builder, therefore no way that it can be used to intercept the request for an already provided client instance. 
Regarding the refactoring for moving the common sdk client initialization in the core module, do we still need an RFC for that? Also, if we are implementing https://github.com/aws-powertools/powertools-lambda-java/issues/1092, maybe it could be part of that?

Issue #1259, opened 2023-06-30
RFC: Handling Large messages for SQS and SNS 


jeromevdl:
## Key information

* RFC PR: (leave this empty)
* Related issue(s), if known:  #596
* Area: SQS
* Meet [tenets](https://docs.powertools.aws.dev/lambda-java/#tenets): yes

## Summary
[summary]: #summary

We should handle large messages (with s3) for [SQS](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-s3-messages.html) and [SNS](https://docs.aws.amazon.com/sns/latest/dg/large-message-payloads.html) and provide a better user experience (error handling, batch...) than today with the `@SqsLargeMessage` annotation.

## Motivation
[motivation]: #motivation

- Lambda Powertools Java does not handle large message for SNS ([doc](https://docs.aws.amazon.com/sns/latest/dg/large-message-payloads.html), [code](https://github.com/aws-powertools/powertools-lambda-java/blob/716a798a5207694ca934857b82279e7e94cc9b2c/powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/internal/SqsLargeMessageAspect.java#L131))
- Current annotation is global to the handler and handles all messages in a batch ([code](https://github.com/aws-powertools/powertools-lambda-java/blob/716a798a5207694ca934857b82279e7e94cc9b2c/powertools-sqs/src/main/java/software/amazon/lambda/powertools/sqs/internal/SqsLargeMessageAspect.java#L67)). Error handling is thus global. It would be better to have the annotation at the item level. See #596.
- Large message handling is currently mixed up with batch processing. While they can be use jointly, large message handling is not needed in all batch cases. As it requires additional libraries (payloadoffloading-common, s3 sdk), it should be an independent module we add specifically for this purpose.

## Proposal
[proposal]: #proposal

### SQS
This feature should be considered in conjunction with #797 but should be in an individual module (eg. `powertools-large-messages` ?). The aim of #797 is to provide a simple interface so that developers can focus on handling each message of the batch while the library takes care of failures. This module should permit to handle large messages individually.

if #797 is providing the following interface:
```java
public void processMessage(SQSMessage message, Context context) {
}
```

[...]

PR #1310, opened 2023-07-25
feat: large message in SQS and SNS

jeromevdl:
**Issue #, if available:** #1259

## Description of changes:

New module to handle large messages in SQS & SNS offloaded to S3.

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

Deprecate part of the SQS module `@SqsLargeMessage` (still available in v1 but will be removed in v2).

**RFC issue #**: 

* [x] Migration process documented
* [x] Implement warnings (if it can live side by side)


powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/LargeMessageConfig.java
```diff
@@ -0,0 +1,59 @@
...
+
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.S3ClientBuilder;
+
+import static software.amazon.lambda.powertools.core.internal.LambdaConstants.AWS_REGION_ENV;
+
+/**
+ * Singleton instance for Large Message Config.
```
scottgerring, line 11: Just to make it super explicit _why singleton_ ... 

```suggestion
 * Singleton instance for Large Message Config. We need this to provide a way to customize the S3 client configuration used by the annotation.
```

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/LargeMessageConfig.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public class LargeMessageConfig {
+
+    private static final LargeMessageConfig INSTANCE = new LargeMessageConfig();
+    private S3Client s3Client;
+
+    private LargeMessageConfig() {
+    }
+
+    public static LargeMessageConfig get() {
```
scottgerring, line 30: I think we should have javadoc on every public interface in the powertools API. WDYT?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/LargeMessageConfig.java
```diff
@@ -0,0 +1,59 @@
...
+        return INSTANCE;
+    }
+
+    public static LargeMessageConfig init() {
+        return INSTANCE;
+    }
+
+    public void withS3Client(S3Client s3Client) {
```
scottgerring, line 38: Should this be `setS3Client`?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/LargeMessageProcessingException.java
```diff
@@ -0,0 +1,11 @@
+package software.amazon.lambda.powertools.largemessages;
+
+public class LargeMessageProcessingException extends RuntimeException {
```
scottgerring, line 3: Perhaps some comments on when we might expect this to be thrown?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageAspect.java
```diff
@@ -0,0 +1,48 @@
...
+ * Handle {@link LargeMessage} annotations.
+ */
+@Aspect
+public class LargeMessageAspect {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LargeMessageAspect.class);
+
+    @SuppressWarnings({"EmptyMethod"})
+    @Pointcut("@annotation(largeMessage)")
+    public void callAt(LargeMessage largeMessage) {
```
scottgerring, line 23: Do we need this at all if we're not using it - is it needed because of the reference on :26?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+            return pjp.proceed(proceedArgs);
+        }
+        // legacy attribute (sqs only)
+        payloadPointer = payloadPointer.replace("com.amazon.sqs.javamessaging.MessageS3Pointer", "software.amazon.payloadoffloading.PayloadS3Pointer");
+
+        LOG.info("Large message [{}]: retrieving content from S3", getMessageId(message));
+
+        String s3ObjectContent = getS3ObjectContent(payloadPointer);
+
+        LOG.debug("Large message [{}]: {}", getMessageId(message), s3ObjectContent);
```
scottgerring, line 50: Is it a good idea to dump `s3ObjectContent` to the log - could be huge?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+import static java.lang.String.format;
+
+/**
+ * Abstract processor for Large Messages. Handle the download from S3 and replace the actual S3 pointer with the content
+ * of the S3 Object leveraging the payloadoffloading library.
+ *
+ * @param <T> any message type that support Large Messages with S3 pointers
+ *            ({@link com.amazonaws.services.lambda.runtime.events.SQSEvent.SQSMessage} and {@link com.amazonaws.services.lambda.runtime.events.SNSEvent.SNSRecord} at the moment)
+ */
+public abstract class LargeMessageProcessor<T> {
```
scottgerring, line 22: I think this should be package private if possible. If it's public we can't change the API to e.g. add another message processor in the future without a breaking change.

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessorFactory.java
```diff
@@ -0,0 +1,23 @@
+package software.amazon.lambda.powertools.largemessages.internal;
+
+import com.amazonaws.services.lambda.runtime.events.SNSEvent.SNSRecord;
+import com.amazonaws.services.lambda.runtime.events.SQSEvent.SQSMessage;
+
+import java.util.Optional;
+
+public class LargeMessageProcessorFactory {
```
scottgerring, line 8: Doesn't need to be public either?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSNSMessageProcessor.java
```diff
@@ -0,0 +1,39 @@
+package software.amazon.lambda.powertools.largemessages.internal;
+
+import com.amazonaws.services.lambda.runtime.events.SNSEvent.MessageAttribute;
+import com.amazonaws.services.lambda.runtime.events.SNSEvent.SNSRecord;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class LargeSNSMessageProcessor extends LargeMessageProcessor<SNSRecord> {
```
scottgerring, line 9: Doesn't need to be public?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSQSMessageProcessor.java
```diff
@@ -0,0 +1,160 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class LargeSQSMessageProcessor extends LargeMessageProcessor<SQSMessage> {
```
scottgerring, line 20: Doesn't need to be public?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+    protected static final String RESERVED_ATTRIBUTE_NAME = "ExtendedPayloadSize";
```
scottgerring, line 23: Is this key available somewhere from within `payloadoffloading-common` ? Seems like something that all the users of the large messaging pattern need to agree upon

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSNSMessageProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    protected String getMessageContent(SNSRecord message) {
+        return message.getSNS().getMessage();
+    }
+
+    @Override
+    protected void updateMessageContent(SNSRecord message, String messageContent) {
+        message.getSNS().setMessage(messageContent);
```
scottgerring, line 23: Do we have to update a hash here too?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSQSMessageProcessor.java
```diff
@@ -0,0 +1,160 @@
...
+     * package protected for testing purpose.
+     *
+     * @param messageBody body of the SQS Message
+     * @return the MD5 digest of the SQS Message body (or empty in case of error)
+     */
+    static Optional<String> calculateMessageBodyMd5(String messageBody) {
+        byte[] expectedMd5;
+        try {
+            expectedMd5 = Md5Utils.computeMD5Hash(messageBody.getBytes(StandardCharsets.UTF_8));
+        } catch (Exception e) {
```
scottgerring, line 76: In what cases might this fail? I'm not sure if silently setting a null digest (and letting the client of powertools maybe have to debug into here to see what's happening) is better than throwing here 🧐

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSQSMessageProcessor.java
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                } else if (attrValue.getBinaryListValues() != null &&
+                        attrValue.getBinaryListValues().size() > 0) {
+                    md5Digest.update(BINARY_LIST_TYPE_FIELD_INDEX);
+                    for (ByteBuffer byteListMember : attrValue.getBinaryListValues()) {
+                        updateLengthAndBytes(md5Digest, byteListMember);
+                    }
+                }
+            }
+        } catch (Exception e) {
```
scottgerring, line 130: Same as above - do we really want to swallow and silently null if we fail to do this?

powertools-sqs/pom.xml
```diff
@@ -15,6 +15,7 @@
 
     <name>Powertools for AWS Lambda (Java) library SQS</name>
     <description>
+        Deprecated: Batch processing is now handled in powertools-batch and large messages in powertools-large-messages modules.
```
scottgerring, line 18: We're going to have to merge batch before we do our next release too :)

scottgerring (commented):
Some initial comments! I think the API is solid and as we discussed on the RFC.

pom.xml
```diff
@@ -75,7 +76,7 @@
         <maven-javadoc-plugin.version>3.5.0</maven-javadoc-plugin.version>
         <maven-source-plugin.version>3.3.0</maven-source-plugin.version>
         <maven-gpg-plugin.version>3.1.0</maven-gpg-plugin.version>
-        <junit-jupiter.version>5.9.3</junit-jupiter.version>
+        <junit-jupiter.version>5.7.2</junit-jupiter.version>
```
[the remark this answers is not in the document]
jeromevdl: because of this: https://github.com/aws-powertools/powertools-lambda-java/pull/1309#issuecomment-1650055388 (tests were not detected in intelliJ...)

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageAspect.java
jeromevdl: yes, we need it, this is how it works...

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessor.java
jeromevdl: why couldn't we add another processor later ? But I must admit that it's in the internal folder so maybe we should keep it really internal...

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessor.java
jeromevdl: Almost, it is in the extended client library (sns and sqs), look at [this](https://github.com/awslabs/amazon-sqs-java-extended-client-lib/blob/master/src/main/java/com/amazon/sqs/javamessaging/SQSExtendedClientConstants.java#L20-L22).

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessor.java
jeromevdl: it's in debug... if they want to debug. 

Should we truncate and log the first 50 char and latest 50 chars ?

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessorFactory.java
jeromevdl: For sure this one does not

jeromevdl:
```suggestion
class LargeMessageProcessorFactory {
```

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSNSMessageProcessor.java
jeromevdl: initially I wanted to be able to use them out of the annotation context, but I've abandoned, so I guess we can put it package
```suggestion
class LargeSNSMessageProcessor extends LargeMessageProcessor<SNSRecord> {
```

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSNSMessageProcessor.java
jeromevdl: not in SNS... there's no hash

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSQSMessageProcessor.java
jeromevdl:
```suggestion
class LargeSQSMessageProcessor extends LargeMessageProcessor<SQSMessage> {
```

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSQSMessageProcessor.java
jeromevdl: It can throw an exception, we must deal with it...  I'm not throwing anything, just returning an empty string

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSQSMessageProcessor.java
jeromevdl: I didn't want to return null, so returned empty but it's the same in the end...

powertools-sqs/pom.xml
jeromevdl: sure, next release will contain batch and large message "v2ish"

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeSQSMessageProcessor.java
scottgerring: What happens when the user ends up with an empty MD5 on their message? Isn't it better to wrap in runtime and throw? This seems _exceptional_ - the user probably wants to see a failure, rather than e.g. comparing body to hash and gettting a mismatch -- ?

jeromevdl: It won't be empty, because if it's empty I put the original MD5. But I agree MD5 won't match if he tries to check (but will he be able to actually check if MD5 is not working for us, there are chances it won't for him..)
I didn't want to raise a exception as I'm pretty sure not so many people are verifying integrity of the messages/attributes...

docs/utilities/large_messages.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Features
+
+- Automatically retrieve the content of S3 objects when SQS or SNS messages have been offloaded to S3.
+- Automatically delete the S3 Objects after processing is OK.
+- Compatible with the batch module (with SQS).
+
+## Background
+
+SQS and SNS message payload is limited to 256KB. If you wish to send larger message payload, you can leverage the
```
scottgerring, line 23:
```suggestion
SQS and SNS message payload is limited to 256KB. If you wish to send messages with a larger payload, you can leverage the
```

docs/utilities/large_messages.md
```diff
@@ -0,0 +1,322 @@
...
+or [amazon-sns-java-extended-client-lib](https://github.com/awslabs/amazon-sns-java-extended-client-lib) which
+offload the message to Amazon S3. See documentation
+([SQS](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-s3-messages.html)
+/ [SNS](https://docs.aws.amazon.com/sns/latest/dg/large-message-payloads.html))
+
+When offloaded to S3, the message contains a specific message attribute and the payload only contains a pointer to the
+S3 object (bucket and object key).
+
+This utility automatically retrieves messages which have been offloaded to S3 using the
+extended client libraries. Once the message payloads have been processed successful, the
```
scottgerring, line 34:
```suggestion
extended client libraries. Once a message's payload has been processed successfully, the
```

docs/utilities/large_messages.md
```diff
@@ -0,0 +1,322 @@
...
+utility also delete the message payloads from S3.
```
scottgerring, line 35:
```suggestion
utility deletes the payload from S3.
```

docs/utilities/large_messages.md
```diff
@@ -0,0 +1,322 @@
...
+
+        @LargeMessage
+        private void processSNSRecord(SNSEvent.SNSRecord snsRecord) {
+            // snsRecord.getSNS().getMessage() will contain the content of the S3 object
+        }
+    }
+    ```
+
+## Migration from the SQS Large Message utility
+
```
scottgerring, line 315: Maybe add something saying "If you are still using the powertools-sqs library for batch processing, consider moving to `powertools-batch` at the same time to remove the dependency on this library completely; it has been deprecated will be removed in v2"

scottgerring (changes requested):
Some small doco nitpicks!

pom.xml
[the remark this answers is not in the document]
scottgerring: Shall we chuck a comment on it so we know not to roll it forward blindly in future?

docs/utilities/large_messages.md
```diff
@@ -0,0 +1,380 @@
...
+- Replace the dependency in maven / gradle: `powertools-sqs` ==> `powertools-large-messages`
+- Replace the annotation: `@SqsLargeMessage` ==> `@LargeMessage` (the new module handles both SQS and SNS)
+- Move the annotation away from the Lambda `handleRequest` method and put it on a method with `SQSEvent.SQSMessage`
+  or `SNSEvent.SNSRecord` as first parameter.
+- The annotation now handles a single message, contrary to the previous version that was handling the complete batch.
+  It gives more control, especially when dealing with partial failures with SQS (see the batch module).
+- The new module only provides an annotation, the `SqsUtils` is not available anymore in this new version.
```
jeromevdl, line 380:
```suggestion
- The new module only provides an annotation, the `SqsUtils` is not available anymore in this new version.

Also, if you are still using the `powertools-sqs` library for batch processing, consider moving to `powertools-batch` at the same time to remove the dependency on this library completely; it has been deprecated and will be removed in v2.
```

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/LargeMessageConfig.java
jeromevdl: as it's a singleton, I prefer with than set (set let me think I could change, with is more a builder thing so only once)

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/LargeMessageProcessingException.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package software.amazon.lambda.powertools.largemessages;
+
+public class LargeMessageProcessingException extends RuntimeException {
```
jeromevdl, line 17:
```suggestion
/**
 * Exception that occurs when the utility fails to retrieve the content from S3
 */
public class LargeMessageProcessingException extends RuntimeException {
```

docs/utilities/large_messages.md
```diff
@@ -7,20 +7,78 @@ The large message utility handles SQS and SNS messages which have had their payl
 offloaded to S3 if they are larger than the maximum allowed size (256 KB).
 
 !!! Notice
-The large message utility (available in the `powertools-batch` module with v1.16.1 or lower) is now deprecated
-and replaced by the `powertools-large-messages` described in this page.
-You can still get the documentation [here](sqs_large_message_handling.md)
-and the migration guide [here](#migration-from-the-sqs-large-message-utility).
+    The large message utility (available in the `powertools-batch` module with v1.16.1 or lower) is now deprecated
```
scottgerring, line 10:
```suggestion
    The large message utility (available in the `powertools-sqs` module with v1.16.1 or lower) is now deprecated
```

docs/utilities/large_messages.md
```diff
@@ -7,20 +7,78 @@ The large message utility handles SQS and SNS messages which have had their payl
...
 
 - Automatically retrieve the content of S3 objects when SQS or SNS messages have been offloaded to S3.
-- Automatically delete the S3 Objects after processing is OK.
+- Automatically delete the S3 Objects after processing succeeds.
 - Compatible with the batch module (with SQS).
 
 ## Background
 
-SQS and SNS message payload is limited to 256KB. If you wish to send larger message payload, you can leverage the
+```mermaid
```
scottgerring, line 23: 🎉

docs/utilities/large_messages.md
```diff
@@ -274,23 +332,62 @@ This utility works perfectly together with the batch module (`powertools-batch`)
     }
     ```
 
+!!! tip "Use together with idempotency module"
+    This utility also works together with the idempotency module (`powertools-idempotency`). 
+    You can add both the `@LargeMessage` and `@Idempotent` annotations to the same method. 
+    The `@Idempotent` takes precedence over the `@LargeMessage` annotation. 
+    It means Idempotency module will use the initial raw message (containing the S3 pointer) and not the large message.
+    Using the large message would end up with potential issues when inserting the data in DynamoDB, where items
```
scottgerring, line 340: Do we need this sentence at all, now that you have organised the precedence?

docs/utilities/large_messages.md
```diff
@@ -319,4 +416,6 @@ import software.amazon.lambda.powertools.largemessages.LargeMessage;
   or `SNSEvent.SNSRecord` as first parameter.
 - The annotation now handles a single message, contrary to the previous version that was handling the complete batch.
   It gives more control, especially when dealing with partial failures with SQS (see the batch module).
-- The new module only provides an annotation, the `SqsUtils` is not available anymore in this new version.
\ No newline at end of file
+- The new module only provides an annotation, the `SqsUtils` is not available anymore in this new version.
```
scottgerring, line 419:
```suggestion
- The new module only provides an annotation, an equivalent to the `SqsUtils` class is not available anymore in this new version.
```

docs/utilities/large_messages.md
```diff
@@ -319,4 +416,6 @@ import software.amazon.lambda.powertools.largemessages.LargeMessage;
...
+
+Also, if you are still using the `powertools-sqs` library for batch processing, consider moving to `powertools-batch` at the same time to remove the dependency on this library completely; it has been deprecated and will be removed in v2.
```
scottgerring, line 421:
```suggestion
Finally, if you are still using the `powertools-sqs` library for batch processing, consider moving to `powertools-batch` at the same time to remove the dependency on this library completely; it has been deprecated and will be removed in v2.
```

pom.xml
```diff
@@ -547,9 +536,9 @@
             </build>
         </profile>
         <profile>
-            <id>jdk11</id>
+            <id>jdk8plus</id>
```
scottgerring, line 539: maybe this is better - `jdk8plus` makes me think jdk8 _and_ newer 
```suggestion
            <id>newerThanJdk8</id>
```

docs/utilities/large_messages.md
jeromevdl: I already say 

> The `@Idempotent` takes precedence over the `@LargeMessage` annotation.

what should we add ?

jeromevdl: maybe something about our discussion and the deleted file in S3...

scottgerring: I think the comment about the precedent is all we need:
> The @Idempotent takes precedence over the @LargeMessage annotation.

This part, explaining (as I understand it) the reason the precedence can't work the other way around, feels unnecessary - with the precedence, you've made sure it doesn't work this way at all, right?

scottgerring: > maybe something about our discussion and the deleted file in S3...

Is this a user-facing detail?

jeromevdl: Not sure to get your comment. But whatever the order you choose for annotations on the method, Idempotency will always be the first executed (and thus using the S3 pointer rather than the full S3 content).
The lifecyle is actually: 
> idempotency (lock the idempotency key) => large message (get S3 content) => function code (`pjp.proceed`) => end of large message code (delete s3 file) => end of idempotency code (store the result in DB)

Regarding the s3 file, I'm actually not sure what can happen:
- If get from S3 or function code fails or deletion fails, the file is not deleted, so we can retry...
- If deletion succeed but end of idempotency fails, ... it's more an idempotency stuff than large message actually. If idempotency fails after the method succeeds, I don't remember what happens but it will be the same here

scottgerring: > But whatever the order you choose for annotations on the method, Idempotency will always be the first executed (and thus using the S3 pointer rather than the full S3 content).

This is what I mean. I think we can stop at line 339:
> It means Idempotency module will use the initial raw message (containing the S3 pointer) and not the large message.

The rest, talking about what would happen if that _wasn't_ the case, is unnecessary detail for the user IMHO.

docs/utilities/large_messages.md
```diff
@@ -0,0 +1,421 @@
...
+        }
+    }
+    ```
+
+!!! tip "Use together with idempotency module"
+    This utility also works together with the idempotency module (`powertools-idempotency`). 
+    You can add both the `@LargeMessage` and `@Idempotent` annotations, in any order, to the same method. 
+    The `@Idempotent` takes precedence over the `@LargeMessage` annotation.
+    It means Idempotency module will use the initial raw message (containing the S3 pointer) and not the large message.
+    Using the large message would end up with potential issues when inserting the data in DynamoDB, where items
```
jeromevdl, line 340:
```suggestion
```

docs/utilities/large_messages.md
```diff
@@ -0,0 +1,419 @@
+---
+title: Large Messages
+description: Utility
+---
+
+The large message utility handles SQS and SNS messages which have had their payloads
+offloaded to S3 if they are larger than the maximum allowed size (256 KB).
+
+!!! Notice
+    The large message utility (available in the `powertools-sqs` module with v1.16.1 or lower) is now deprecated
```
jeromevdl, line 10:
```suggestion
    The large message utility (available in the `powertools-sqs` module for versions v1.16.1 and earlier) is now deprecated
```

scottgerring (approved):
LGTM!

Issue #1076, opened 2023-02-24
Maintenance: v2 - rename 'core' module 


scottgerring:
### Summary

The 'powertools-core' module provides common functionality to the other powertools modules, and does not have anything to do with the externally-facing idea of 'powertools core functionality' [logging/tracing/metrics]./ 

I propose that we rename this to something else (common? shared? 🤷 ) as part of V2.

### Why is this needed?

It's a confusing detail where our interface - the modules we expose - doesn't match the documentation in a strange fashion. You'd be forgiven for thinking if you ask for powertools-core you get all of logging/tracing/metrics.

### Which area does this relate to?

Other

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-java/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

YuvrajPraveenSingh:
O would like to work on this if possible

jeromevdl:
@YuvrajPraveenSingh thanks for your support. However this is something we'll do on a v2 and therefore not right now.

PR #1364, opened 2023-08-15
chore: [V2] rename 'core' module to 'common'

scottgerring:
**Issue #, if available:**
#1076 

## Description of changes:
The Java 'core' module has nothing to do with the powertools concept of 'core' utilities - logging, tracing, metrics - but is rather common functionality shared by the other PT modules.

To clarify this distinction, we will rename this to 'powertools-common', which fits with the other PT libraries naming scheme.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


powertools-common/src/main/java/software/amazon/lambda/powertools/core/internal/LambdaConstants.java
jeromevdl, line 1: Package should be changed too

scottgerring:
We still have `examples/powertools-examples-core` . Previously this was super wrong - it had nothing to do with the 'core' module.

Now it is a little less wrong - it is the concept of "core utilities", which are split over `powertools-logging`, `powertools-tracing`, `powertools-metrics`. 

What should we call the example?

scottgerring:
Fixed namespace

powertools-common/pom.xml
```diff
@@ -27,7 +27,7 @@
         <version>2.0.0-SNAPSHOT</version>
     </parent>
 
-    <name>Powertools for AWS Lambda (Java) library Core</name>
+    <name>Powertools for AWS Lambda (Java) library Common Utilities</name>
     <description>
         A suite of utilities for AWS Lambda Functions that makes tracing with AWS X-Ray, structured logging and creating custom metrics asynchronously easier.
```
jeromevdl, line 32: All the content of common is internal, maybe we could simplify this description by just saying common internal utilities...

jeromevdl (commented):
couple of comments

powertools-common/pom.xml
```diff
@@ -27,10 +27,8 @@
         <version>2.0.0-SNAPSHOT</version>
     </parent>
 
-    <name>Powertools for AWS Lambda (Java) library Core</name>
-    <description>
-        A suite of utilities for AWS Lambda Functions that makes tracing with AWS X-Ray, structured logging and creating custom metrics asynchronously easier.
-    </description>
+    <name>Powertools for AWS Lambda (Java) library Common Internal Utilities</name>
```
scottgerring, line 30: @jeromevdl how about this (and description)

jeromevdl: I would change description to something like

> Internal utilities used in Lambda Powertools modules, do not use it directly in your project

scottgerring: Done

jeromevdl (approved):
LGTM !

Issue #1239, opened 2023-06-23
Maintenance: Version the documentation


jeromevdl:
### Summary

Currently the documentation is not versioned, unlike [Python](https://docs.powertools.aws.dev/lambda-python/2.17.0/) and [Typescript](https://docs.powertools.aws.dev/lambda-typescript/latest/). We should version the doc to be consistent with other runtimes and to prepare for v2.

### Why is this needed?

- Consistency with other runtimes
- Keep history of changes (doc for a specified version)
- Preparing v2 (we'll need to have both v1 and v2 available at least)

### Which area does this relate to?

Automation, Governance

### Solution

See how python is doing: https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/workflows/reusable_publish_docs.yml. Check with @sthulb / @heitorlessa.

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda (Java) languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

sthulb:
Python/TS uses a python package called `mike` to version docs – but in our post GH Pages world, we might want to look into doing this differently.

kozub:
I'm going to take care of it.

dreamorosi:
Ok so I've been taking a look at this and I have a proposal.

From an MkDocs standpoint, all we need to do to enable the versioning is to add these lines in the `mkdocs.yml` file under `extra`:

```yml
extra:
  version:
    provider: mike
    default: latest # the is the alias we want to be the main
```

Whenever this is enabled, MkDocs will try to load a `versions.json` file from the root of the static site, if present it'll use it to render a dropdown with the versions at the top of the site:

![Image](https://github.com/user-attachments/assets/7b6d67cc-8db9-42c7-8789-b4ad6b0b9c64)

The `versions.json` file should have a structure like the one below, and despipe MkDocs thinking it's built using [`mike`](https://pypi.org/project/mike/), it can be built manually. We know this works because we've ditched `mike` over a year ago in TypeScript and Python and still have versioning.

```json
[
  {
    "title": "2 (preview)",
    "version": "2.x",
    "aliases": [
      "latest"
    ]
  },
  {
    "title": "legacy",
[...]

PR #1370, opened 2023-08-16
docs: Add support for docs versioning (#1239)

kozub:
**Issue #, if available:**
#1239

## Description of changes:
Add support for docs versioning and publishing docs to S3 with version.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs
* [] PR title follows [conventional commit semantics](https://www.conventionalcommits.org/en/v1.0.0/)


.github/workflows/release-doc.yml
```diff
@@ -0,0 +1,19 @@
+name: Docs
```
scottgerring, line 1: This is the "normal" one that will get triggered and deploy to S3 automatically when we publish a release, right?

.github/workflows/reusable-publish-docs.yml
```diff
@@ -0,0 +1,118 @@
...
+
+jobs:
+  publish-docs:
+    runs-on: ubuntu-latest
+    environment: Docs
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab  # v3.5.2
+        with:
+          # While `fetch-depth` is used to allow the workflow to later commit & push the changes.
```
scottgerring, line 38: I don't think we push anything

.github/workflows/release-doc.yml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch: {}
+
+jobs:
+  release-docs:
+    permissions:
+      contents: write
```
scottgerring, line 12: We're not doing github pages anymore, right? So we can drop `pages`. I don't believe we're pushing docs back to the repo either, so we can drop `contents`

.github/workflows/reusable-publish-docs.yml
```diff
@@ -0,0 +1,118 @@
...
+        required: true
+        type: string
+      detached_mode:
+        description: "Whether it's running in git detached mode to ensure git is sync'd"
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: write
```
scottgerring, line 26: Same as above - I don't think we need `contents` or `pages`

.github/workflows/reusable-publish-docs.yml
```diff
@@ -0,0 +1,118 @@
...
+          git pull origin "$BRANCH"
+      - name: Normalize Version Number
+        run: echo "VERSION=$(echo ${{ inputs.version }} | sed 's/v//')" >> $GITHUB_ENV
+      - name: Build docs website and API reference
+        env:
+          ALIAS: ${{ inputs.alias }}
+        run: |
+          rm -rf site
+          mkdocs build
+          mike deploy --update-aliases --no-redirect ${{ env.VERSION }} ${{ env.ALIAS }} --branch backup-gh-pages
```
scottgerring, line 67: I gather we're doing this to pretend we're deploying to pages. Is this step actually trying to "push" ?

.github/workflows/reusable-publish-docs.yml
```diff
@@ -0,0 +1,118 @@
...
+        uses: actions/setup-python@57ded4d7d5e986d7296eab16560982c6dd7c923b # v4.6.0
+        with:
+          python-version: "3.8"
+      - name: Install doc generation dependencies
+        run: |
+          pip install --upgrade pip 
+          pip install -r docs/requirements.txt
+      - name: Setup doc deploy
+        run: |
+          git config --global user.name Docs deploy
```
scottgerring, line 50: If we're not committing and pushing, we don't need to set the user.email or user.name

.github/workflows/rebuild-latest-docs.yml
```diff
@@ -0,0 +1,39 @@
+name: Rebuild latest docs
+
+# PROCESS
+#
+# 1. Build User Guide and API docs
+# 2. Publish to GitHub Pages
```
scottgerring, line 6: I don't think we need this?

mkdocs.yml
```diff
@@ -13,16 +13,11 @@ nav:
   - Utilities:
       - utilities/idempotency.md
       - utilities/parameters.md
-      - utilities/large_messages.md
+      - utilities/sqs_large_message_handling.md
```
scottgerring, line 16: These changes need to be rolled back -  from L16-L25. We released new modules and changed the doc structure after you forked!

scottgerring (changes requested):
Almost there I think! Wondering if we need to keep the gh-pages cruft.

.github/workflows/release-doc.yml
```diff
@@ -0,0 +1,19 @@
...
+      pages: write
+      id-token: write
+    secrets: inherit
+    uses: ./.github/workflows/reusable-publish-docs.yml
+    with:
+      version: main
```
kozub, line 18: Just found that version here is hardcoded. I need to check how it use here released version instead of "main".

.github/workflows/rebuild-latest-docs.yml
kozub: it just describes steps of the process which is done here (except "and API docs" which can be removed). I think it might be helpful to understand what we do here.

.github/workflows/reusable-publish-docs.yml
kozub: Thanks actually what it do: "build docs and deploy them to a branch".

.github/workflows/reusable-publish-docs.yml
kozub: There is a separate branch `gh-pages` for storing versioned docs as backup and we commit and push to it.

.github/workflows/reusable-publish-docs.yml
scottgerring: My bad. And I see the other repos all keep this, too

.github/workflows/reusable-publish-docs.yml
scottgerring: Resolving as we keep `gh-pages`

.github/workflows/reusable-publish-docs.yml
scottgerring: Based on `gh-pages` - probably we need `contents` but not `pages` ?

.github/workflows/release-doc.yml
scottgerring: Based on gh-pages - probably we need `contents` but not `pages` ?

.github/workflows/rebuild-latest-docs.yml
scottgerring: The step "# 2. Publish to GitHub Pages" - we don't need this anymore, do we - maybe it becomes "push versioned docs to gh-pages branch" or something?

.github/workflows/release-doc.yml
scottgerring: Looking at `reusable-publish-docs.yml`, it seems like it should be the actual version alias, not a git ref? It looks like it is dropping the files into there in S3 and using it for the version label 🧐 

https://github.com/aws-powertools/powertools-lambda-java/blob/910085907687a565d04fc32303e1159963671bac/.github/workflows/reusable-publish-docs.yml#L79-L86

scottgerring:
Let's close this for now and loop back when we have capacity @kozub and @sthulb

PR #1388, opened 2023-08-21
chore: skip deployment in tests and examples

jeromevdl:
**Issue #, if available:**

## Description of changes:

Replaced the property `maven.deploy.skip` with the full plugin configuration to avoid deployment of tests and examples :

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


scottgerring:
Perhaps this would work better the other way around - opt-in, rather than opt-out? Going forward we're likely to add more examples than modules I suspect, and that way the failure mode isn't accidentally publishing things.

jeromevdl:
> Perhaps this would work better the other way around - opt-in, rather than opt-out? Going forward we're likely to add more examples than modules I suspect, and that way the failure mode isn't accidentally publishing things. What do you think @jeromevdl ?

I don't know how/if it will work if we put the skip true on top and false below...

scottgerring:
> > Perhaps this would work better the other way around - opt-in, rather than opt-out? Going forward we're likely to add more examples than modules I suspect, and that way the failure mode isn't accidentally publishing things. What do you think @jeromevdl ?
> 
> I don't know how/if it will work if we put the skip true on top and false below...

Neither do I. The other question is - is there a way we can be sure this will work, this way around, without waiting til the next release?

scottgerring (approved):

Issue #1402, opened 2023-08-29
Maintenance: Split powertools-parameters into submodules


scottgerring:
### Summary

In order to minimize dependency size, we should split `powertools-parameters` into submodules (keeping `powertools-parameters` as the interface) :

- powertools-parameters-ssm
- powertools-parameters-secrets
- powertools-parameters-dynamodb
- powertools-parameters-appconfig

This is a breaking change and should be done in v2 only.

### Why is this needed?

This will allow users to only pull in dependencies that they need for their app, rather then our extra code plus the pieces of the SDK that needs. 

### Which area does this relate to?

Parameters

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda (Java) languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

jeromevdl:
done with #1403, will be released with v2

PR #1403, opened 2023-08-29
chore(v2): Split parameters module up by parameter provider

scottgerring:
**Issue #, if available:**
#1402 

## Description of changes:

**Checklist**

 

* [x] Update tests
* [x] Update docs

**Outstanding Changes**

* [ ] Migration guide
* [ ] Documentation
* [ ] AspectJ test coverage

## Breaking change checklist

**RFC issue #**:


powertools-parameters/powertools-parameters-appconfig/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>software.amazon.lambda</groupId>
+        <artifactId>powertools-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
```
scottgerring, line 11: Then we have one of these per-provider, stored in the `powertools-parameters` subdirectory, pulled into the entire repo parent POM.

@jeromevdl what do you think - should we nest this at all, or just have all the modules at the top directory? I am now leaning towards the latter - some stuff (like the relative path to `spotbugs-exclude.xml`) is easier when all of the deployable modules are one level deep.

powertools-parameters/powertools-parameters-tests/pom.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
scottgerring, line 1: Some of the tests cut across multiple modules, and have been moved into their own project to facilitate this.

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbParam.java
```diff
@@ -0,0 +1,31 @@
...
+
+package software.amazon.lambda.powertools.parameters.dynamodb;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import software.amazon.lambda.powertools.parameters.transform.Transformer;
+
+@Retention(RetentionPolicy.RUNTIME)
```
scottgerring, line 23: Each provider now has its own annotation. This works better IMHO because 1/ it makes the module dependencies simpler and 2/ some parameters which you couldn't inject before because of the custom arguments - e.g. DynamoDb - can now be injected.

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbParamAspect.java
```diff
@@ -12,43 +12,45 @@
...
-import software.amazon.lambda.powertools.parameters.ParamManager;
 
 @Aspect
-public class LambdaParametersAspect {
+public class DynamoDbParamAspect {
 
-    @Pointcut("get(* *) && @annotation(paramAnnotation)")
-    public void getParam(Param paramAnnotation) {
+
+    @Pointcut("get(* *) && @annotation(ddbConfigParam)")
```
scottgerring, line 27: ... But this also means we have some duplication of the aspect itself. I will make sure we have test coverage of each of these, as there is enough "compile time magic" that it's nice to cover that the aspect works.

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbProviderBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.lambda.powertools.common.internal.UserAgentConfigurator;
+import software.amazon.lambda.powertools.parameters.BaseProvider;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
+
+class DynamoDbProviderBuilder {
```
scottgerring, line 28: I've moved all of the builders out into their own classes. Before we had a weird mix of places that configured the clients --> `ParamManager` class, each `XParamProvider`, and each `XParamProviderBuilder`. This made it quite hard to reason about what was being configured where and why. Now we have one place, and it is the builder.

scottgerring:
@jeromevdl when you have a moment, I would love some initial feedback here! It is not ready for merge - docs and tests need updating - but I have fleshed out the structure of it. Github comments guide you through my reasoning.

powertools-parameters/powertools-parameters-appconfig/pom.xml
jeromevdl: I agree it's probably easier to keep it on root, but I think we start to have too many things at the root level. Would be cleaner to have it nested. And I should probably do the same for logging. Have one logging and then the log4j / logback implementations nested.

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbParamAspect.java
jeromevdl: I don't know what is best here... 

Having different annotations means users will have to specify the good project in their weaving configuration in maven/gradle. We already have tons of questions/issues because of this.

But it's also probably cleaner (despite the duplication) for us and opens more doors (different config based on the store used). It's also maybe easier to read for users, more explicit: @DynamoDbParam, @SecretParam, ...

Let's go that way and document it / test it properly.

examples/powertools-examples-core-utilities/gradle/README.md
jeromevdl, line 1: why do we have this here ??

jeromevdl: all the gradle example should not be in that PR...

examples/powertools-examples-parameters/src/main/java/org/demo/parameters/ParametersFunction.java
```diff
@@ -31,15 +31,28 @@
...
-    SecretsProvider secretsProvider = ParamManager.getSecretsProvider();
+    // Annotation-style injection from secrets manager
+    @SecretsParam(key = "/powertools-java/userpwd")
+    String secretParamInjected;
+
+    // Annotation-style injection from Systems Manager
+    @SSMParam(key = "/powertools-java/sample/simplekey")
+    String ssmParamInjected;
+
+    SSMProvider ssmProvider = SSMProvider
```
jeromevdl, line 50: looks weird  `.builder().build()` could we do like the SDK have a `.create()` that does the same?

(same for other providers)

examples/powertools-examples-parameters/src/main/java/org/demo/parameters/ParametersFunction.java
```diff
@@ -31,15 +31,28 @@
...
+            .builder()
+            .build();
+    SecretsProvider secretsProvider = SecretsProvider
+            .builder()
+            .build();
 
     String simpleValue = ssmProvider.defaultMaxAge(30, SECONDS).get("/powertools-java/sample/simplekey");
```
jeromevdl, line 57: maybe defaultMaxAge should be a builder option ?

pom.xml
```diff
@@ -54,6 +54,11 @@
         <module>powertools-e2e-tests</module>
         <module>powertools-batch</module>
         <module>examples</module>
+        <module>powertools-parameters/powertools-parameters-ssm</module>
+        <module>powertools-parameters/powertools-parameters-secrets</module>
+        <module>powertools-parameters/powertools-parameters-dynamodb</module>
+        <module>powertools-parameters/powertools-parameters-tests</module>
```
jeromevdl, line 60: move after appconfig

powertools-large-messages/src/main/java/software/amazon/lambda/powertools/largemessages/internal/LargeMessageProcessor.java
jeromevdl, line 1: why ?

powertools-parameters/powertools-parameters-appconfig/src/main/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigParam.java
```diff
@@ -12,34 +12,22 @@
...
- * The user can implement a custom provider by extending the abstract class {@link BaseProvider}.</p>
- *
- * <p>If the parameter value requires transformation before being assigned to the annotated field
- * users can specify a {@link Transformer}
- * </p>
- */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
-public @interface Param {
+public @interface AppConfigParam {
```
jeromevdl, line 25: comment / description

powertools-parameters/powertools-parameters-appconfig/src/test/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderTest.java
jeromevdl, line 1: why did you remove the static imports ?

powertools-parameters/powertools-parameters-secrets/src/main/java/software/amazon/lambda/powertools/parameters/secrets/SecretsParamAspect.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Around("getParam(secretsParam)")
+    public Object injectParam(final ProceedingJoinPoint joinPoint, final SecretsParam secretsParam) {
+        System.out.println("GET IT");
+
+        SecretsProvider provider = SecretsProvider.builder()
+                .build();
+
+        if (secretsParam.transformer().isInterface()) {
```
jeromevdl, line 38: here again same code

powertools-parameters/powertools-parameters-secrets/src/test/java/software/amazon/lambda/powertools/parameters/secrets/SecretsProviderTest.java
```diff
@@ -82,27 +85,28 @@ public void getValueBase64() {
...
+        Assertions.assertThat(value).isEqualTo(expectedValue);
+        Assertions.assertThat(paramCaptor.getValue().secretId()).isEqualTo(key);
     }
 
     @Test
     public void getMultipleValuesThrowsException() {
 
         // Act & Assert
-        assertThatRuntimeException().isThrownBy(() -> provider.getMultipleValues("path"))
+        Assertions.assertThatRuntimeException().isThrownBy(() -> provider.getMultipleValues("path"))
```
jeromevdl, line 96: why not static ?

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMParamAspect.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Around("getParam(SSMParam)")
+    public Object injectParam(final ProceedingJoinPoint joinPoint, final SSMParam SSMParam) {
+        System.out.println("GET IT");
+
+        SSMProvider provider = SSMProvider.builder()
+                .build();
+
+        if (SSMParam.transformer().isInterface()) {
```
jeromevdl, line 38: :)

powertools-parameters/powertools-parameters-ssm/src/test/java/software/amazon/lambda/powertools/parameters/ssm/SSMProviderTest.java
```diff
@@ -12,27 +12,25 @@
...
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-import static org.mockito.MockitoAnnotations.openMocks;
 
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import org.assertj.core.api.Assertions;
```
jeromevdl, line 24: static imports...

powertools-parameters/spotbugs-exclude.xml
```diff
@@ -0,0 +1,81 @@
+<!-- This file specifies a spotbugs filter for excluding reports that
+     should not be considered errors.
+     The format of this file is documented at:
+       https://spotbugs.readthedocs.io/en/latest/filter.html
+     When possible, please specify the full names of the bug codes,
+     using the pattern attribute, to make it clearer what reports are
+     being suppressed.  You can find a listing of codes at:
+       https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html
+  -->
+<FindBugsFilter>
```
jeromevdl, line 10: are we sure we cannot afford these?

spotbugs-exclude.xml
```diff
@@ -8,6 +8,40 @@
        https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html
   -->
 <FindBugsFilter>
+    <Match>
```
jeromevdl, line 11: duplicate ?

jeromevdl (commented):
I think the structure is OK, maybe there's a piece of code we could try to move to common if possible.

powertools-parameters/powertools-parameters-appconfig/pom.xml
scottgerring: I'll leave it like it is, then

powertools-parameters/powertools-parameters-appconfig/src/test/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderTest.java
scottgerring: I suspect that some part of the code autoformatting did this; I certainly haven't done it by hand.

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbParamAspect.java
scottgerring: I try to _bias for loose coupling_ when things get weird, and this was broken before - the annotation could only inject ~half of the parameter providers, and had weird coupling every which way. Agree that it needs good test coverage; weaving is error prone!

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbParamAspect.java
```diff
@@ -12,43 +12,45 @@
...
+    @Around("getParam(ddbConfigParam)")
+    public Object injectParam(final ProceedingJoinPoint joinPoint, final DynamoDbParam ddbConfigParam) {
+        System.out.println("GET IT");
+
+        DynamoDbProvider provider = DynamoDbProvider.builder()
+                .withTable(ddbConfigParam.table())
+                .build();
 
-        if (paramAnnotation.transformer().isInterface()) {
+        if (ddbConfigParam.transformer().isInterface()) {
```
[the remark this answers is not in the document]
scottgerring: I think so. I will come back to this.

examples/powertools-examples-parameters/src/main/java/org/demo/parameters/ParametersFunction.java
scottgerring: I honestly think `build` is more common? Maybe the issue is that we're using the builder without providing anything?

scottgerring:
I found some other things that I think should be cleaned up. 

**First Thing**
on `BaseProvider`

https://github.com/aws-powertools/powertools-lambda-java/blob/54c16647369d72f3b02e8e90098715f25ebe24a5/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java#L72-L76

https://github.com/aws-powertools/powertools-lambda-java/blob/54c16647369d72f3b02e8e90098715f25ebe24a5/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java#L128-L144

The idea is, you can use the provider interface like this, mutating the provider itself, and then the `get` clears up the mutation. This means we're not threadsafe where we could be, which is probably not a big deal for Lambda, but it's there.

```java
provider..withMaxAge(10, Seconds).get("myKey")
```

Now keeping `withMaxAge` in mind, consider this:

https://github.com/aws-powertools/powertools-lambda-java/blob/54c16647369d72f3b02e8e90098715f25ebe24a5/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java#L72-L76

Here we're 1/ mutating something that _is not reset_ by the `get` but still following the builder-style interface and 2/ mutating a field of another object that we hold, but do not necessarily own. 

The upshot of this is that the interface is not consistent with itself, and users may get confused when a shared CacheManager's behaviour is mutated by a call to a provider directly.

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMProvider.java
```diff
@@ -124,33 +107,6 @@ public String getValue(String key) {
         return client.getParameter(request).parameter().value();
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public SSMProvider defaultMaxAge(int maxAge, ChronoUnit unit) {
```
scottgerring, line 131: I believe this should be mutated on the CacheManager itself, not on this interface. Changing it here breaks encapsulation in a non-intuitive fashion

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMProvider.java
```diff
@@ -124,33 +107,6 @@ public String getValue(String key) {
...
-        super.defaultMaxAge(maxAge, unit);
-        return this;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public SSMProvider withMaxAge(int maxAge, ChronoUnit unit) {
```
scottgerring, line 140: I pulled this up to the base class. The only two transformations you can do here are common to all - `withMaxAge` and `withTransformation` - so there's no need to override and reimplement this everywhere.

I would like to change this so it returns some other object - e.g. `GetQueryBuilder` or so - so we have a natural scope for the mutation of the parameters associated with the get, and don't have to mutate and un-mutate the Baseprovider itself (see `BaseProvider::resetToDefaults::`)

I see this as a subsequent change almost worthy of an RFC and not something to do here - this is already a massive PR. 

```
    protected void resetToDefaults() {
        cacheManager.resetExpirationTime();
        if (transformationManager != null) {
            transformationManager.setTransformer(null);
        }
    }
```

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
```diff
@@ -34,11 +34,12 @@ public abstract class BaseProvider implements ParamProvider {
     public static final String PARAMETERS = "parameters";
```
jeromevdl, line 34: not used, remove ?

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMParamAspect.java
```diff
@@ -0,0 +1,50 @@
...
+    private static Supplier<SSMProvider> providerBuilder = () -> SSMProvider.builder()
+            .build();
+
+    @Pointcut("get(* *) && @annotation(secretsParam)")
+    public void getParam(SSMParam secretsParam) {
+    }
+
+    @Around("getParam(ssmPaam)")
+    public Object injectParam(final ProceedingJoinPoint joinPoint, final SSMParam ssmPaam) {
+        System.out.println("GET IT");
```
jeromevdl, line 43:
```suggestion
```

powertools-parameters/powertools-parameters-secrets/src/main/java/software/amazon/lambda/powertools/parameters/secrets/SecretsProvider.java
```diff
@@ -0,0 +1,108 @@
...
+import software.amazon.lambda.powertools.parameters.BaseProvider;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
+
+/**
+ * AWS Secrets Manager Parameter Provider<br/><br/>
+ *
+ * <u>Samples:</u>
+ * <pre>
+ *     SecretsProvider provider = ParamManager.getSecretsProvider();
```
jeromevdl, line 32: `ParamManager` does not exist anymore

powertools-parameters/spotbugs-exclude.xml
```diff
@@ -0,0 +1,95 @@
...
+    </Match>
+    <Match>
+        <Bug pattern="MS_EXPOSE_REP"/>
+        <Or>
+            <And>
+                <Class name="software.amazon.lambda.powertools.parameters.ParamManager"/>
+                <Method name="getCacheManager"/>
+            </And>
+            <And>
+                <Class name="software.amazon.lambda.powertools.parameters.ParamManager"/>
```
jeromevdl, line 90: `ParamManager` does not exist anymore

powertools-parameters/powertools-parameters-secrets/src/main/java/software/amazon/lambda/powertools/parameters/secrets/SecretsProvider.java
jeromevdl: by the way, it looks like e2e tests have not been udpated (`Function` in parameters handler of the E2E) is stil using the `ParamManager`

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMProvider.java
jeromevdl, line 1: "contained" => contain

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMProviderBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Create a {@link SSMProvider} instance.
+     *
+     * @return a {@link SSMProvider}
+     */
+    public SSMProvider build() {
+        if (cacheManager == null) {
+            // TODO - do we want to share this somehow?
```
jeromevdl, line 50: do we? or do we remove the TODO ?
what do you mean by sharing the cacheManager ?

powertools-parameters/powertools-parameters-appconfig/src/main/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+                            .putAdvancedOption(SdkAdvancedClientOption.USER_AGENT_SUFFIX,
+                                    UserAgentConfigurator.getUserAgent(BaseProvider.PARAMETERS)).build())
+                    .build();
+        }
+
+        return new AppConfigProvider(cacheManager, transformationManager, client, environment, application);
+    }
+
+    /**
+     * Set custom {@link AppConfigProvider} to pass to the {@link AppConfigDataClient}. <br/>
```
jeromevdl, line 66:
```suggestion
     * Set custom {@link AppConfigDataClient} to pass to the {@link AppConfigProvider}. <br/>
```

powertools-parameters/powertools-parameters-appconfig/src/main/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param application the application to pull configuration from
+     * @return the builder to chain calls (eg. <pre>builder.withCacheManager().build()</pre>)
+     */
+    public AppConfigProviderBuilder withApplication(String application) {
+        this.application = application;
+        return this;
+    }
+
+    /**
+     * <b>Mandatory</b>. Provide a CacheManager to the {@link AppConfigProvider}
```
jeromevdl, line 100: this is marked as Mandatory but if null we [create a new one](https://github.com/aws-powertools/powertools-lambda-java/pull/1403/files#diff-b307f4bdf939eae47372ead9752df4e3da4a3d0d4f360bc08619c1a59bf08663R42)

jeromevdl: Same for other providers. Is this really mandatory?

powertools-parameters/powertools-parameters-dynamodb/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+            <organizationUrl>https://aws.amazon.com/</organizationUrl>
+        </developer>
+    </developers>
+
+    <distributionManagement>
+        <snapshotRepository>
+            <id>ossrh</id>
+            <url>https://aws.oss.sonatype.org/content/repositories/snapshots</url>
+        </snapshotRepository>
+    </distributionManagement>
```
jeromevdl, line 38: you can clean this I think ;) and same for other modules

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbProvider.java
```diff
@@ -0,0 +1,121 @@
...
+import software.amazon.awssdk.services.dynamodb.model.QueryRequest;
+import software.amazon.awssdk.services.dynamodb.model.QueryResponse;
+import software.amazon.lambda.powertools.parameters.BaseProvider;
+import software.amazon.lambda.powertools.parameters.ParamProvider;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+import software.amazon.lambda.powertools.parameters.dynamodb.exception.DynamoDbProviderSchemaException;
+import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
+
+/**
+ * Implements a {@link ParamProvider} on top of DynamoDB. The schema of the table
```
jeromevdl, line 33:
```suggestion
 * Implements a {@link ParamProvider} on top of Amazon DynamoDB. The schema of the table
```

powertools-parameters/powertools-parameters-appconfig/src/main/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.appconfigdata.AppConfigDataClient;
+import software.amazon.lambda.powertools.common.internal.UserAgentConfigurator;
+import software.amazon.lambda.powertools.parameters.BaseProvider;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
+
+public class AppConfigProviderBuilder {
```
jeromevdl, line 28: small comment like "Builder for the {@link AppConfigProvider}"

powertools-parameters/powertools-parameters-secrets/src/main/java/software/amazon/lambda/powertools/parameters/secrets/SecretsParam.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import software.amazon.lambda.powertools.parameters.transform.Transformer;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface SecretsParam {
```
jeromevdl, line 25: Maybe worth documenting a bit more these annotation, as it is the main entry point for users.
What it is and a small usage example would be great.

powertools-parameters/powertools-parameters-secrets/src/main/java/software/amazon/lambda/powertools/parameters/secrets/SecretsProviderBuilder.java
```diff
@@ -0,0 +1,97 @@
...
+import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
+import software.amazon.lambda.powertools.common.internal.UserAgentConfigurator;
+import software.amazon.lambda.powertools.parameters.BaseProvider;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
+
+public class SecretsProviderBuilder {
```
jeromevdl, line 28: small comment like "Builder for the {@link SecretsProvider}"

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMParam.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import software.amazon.lambda.powertools.parameters.transform.Transformer;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface SSMParam {
```
jeromevdl, line 25: doc

powertools-parameters/powertools-parameters-ssm/src/main/java/software/amazon/lambda/powertools/parameters/ssm/SSMProviderBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param client Custom client
+     * @return the builder to chain calls (eg. <pre>builder.withClient().build()</pre>)
+     */
+    public SSMProviderBuilder withClient(SsmClient client) {
+        this.client = client;
+        return this;
+    }
+
+    /**
+     * <b>Mandatory</b>. Provide a CacheManager to the {@link SSMProvider}
```
jeromevdl, line 76: Mandatory?

powertools-parameters/powertools-parameters-tests/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+        <dependency>
+            <groupId>software.amazon.lambda</groupId>
+            <artifactId>powertools-parameters-secrets</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.lambda</groupId>
+            <artifactId>powertools-parameters-dynamodb</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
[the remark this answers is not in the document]
jeromevdl: put them all in scope test

powertools-parameters/powertools-parameters-tests/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
```diff
@@ -149,7 +151,6 @@ public void get_customDefaultTTL_cached_shouldGetFromCache() {
 
     @Test
     public void get_customDefaultTTL_expired_shouldGetValue() {
-        provider.defaultMaxAge(2, ChronoUnit.MINUTES).get("barbaz");
```
jeromevdl, line 152: not sure to understand why you removed all the different test config ?

powertools-parameters/powertools-parameters-tests/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
jeromevdl, line 1: I don't understand the new tests. You removed all the different configurations.

powertools-parameters/powertools-parameters-tests/src/test/java/software/amazon/lambda/powertools/parameters/ParamProvidersIntegrationTest.java
```diff
@@ -112,7 +113,9 @@ public void ssmProvider_getMultiple() {
 
     @Test
     public void secretsProvider_get() {
-        SecretsProvider secretsProvider = ParamManager.getSecretsProvider(secretsManagerClient);
+        SecretsProvider secretsProvider = SecretsProvider.builder()
```
jeromevdl, line 116: much better these builders than the old ParamManager

docs/utilities/parameters.md
```diff
@@ -4,29 +4,40 @@ description: Utility
...
 
 
-The parameters utility provides a way to retrieve parameter values from
+The parameters utilities provide a way to retrieve parameter values from
 [AWS Systems Manager Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html), 
 [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/), or [Amazon DynamoDB](https://aws.amazon.com/dynamodb/). 
-It also provides a base class to create your parameter provider implementation.
-
+ 
 **Key features**
```
jeromevdl, line 11:
```suggestion
## Key features
```

docs/utilities/parameters.md
```diff
@@ -4,29 +4,40 @@ description: Utility
...
+* Retrieve one or multiple parameters from an underlying provider in a standard way
 * Cache parameter values for a given amount of time (defaults to 5 seconds)
 * Transform parameter values from JSON or base 64 encoded strings
 
 ## Install
-Depending on your version of Java (either Java 1.8 or 11+), the configuration slightly changes.
+In order to provide lightweight dependencies, each parameters module is available as its own
+package:
+
+* **Secrets Manager** - `powertools-parameters-secrets` - [A service that centrally manages the lifecycle of secrets](https://aws.amazon.com/secrets-manager/). Start here if you need to manage secrets only.
```
jeromevdl, line 21: It's not clear for me these `Start here` ...

jeromevdl: why not Start here for DDB then ?

jeromevdl: I would remove the extra blabla, just references the modules and the underlying service they leverage.

docs/utilities/parameters.md
```diff
@@ -4,29 +4,40 @@ description: Utility
...
+package:
+
+* **Secrets Manager** - `powertools-parameters-secrets` - [A service that centrally manages the lifecycle of secrets](https://aws.amazon.com/secrets-manager/). Start here if you need to manage secrets only. 
+* **SSM Parameter Store** - `powertools-parameters-ssm` - [A secure, hierarchical store for configuration data management](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html). Start here if you need to manage various types of configuration data. 
+* **Amazon DynamoDB** -`powertools-parameters-dynamodb` - [A fast, flexible, NoSQL database](https://aws.amazon.com/dynamodb/)
+* **AWS AppConfig** - `powertools-parameters-appconfig` - [An application focussed configuration store](https://aws.amazon.com/systems-manager/features/appconfig) - start here if you are interested in advanced use cases, such as incremental deployment of configuration changes
+
+You can easily mix and match parameter providers within the same project for different needs.  
+
+Depending on which Java version you are using, you configuration will differ. Note that you must also provide
```
jeromevdl, line 28: this TODO thing is also not super clear, see my proposition below

docs/utilities/parameters.md
```diff
@@ -4,29 +4,40 @@ description: Utility
...
-    ```xml hl_lines="3-7 16 18 24-27"
+    ```xml hl_lines="4-7 16 18 26-28"
     <dependencies>
         ...
         <dependency>
             <groupId>software.amazon.lambda</groupId>
-            <artifactId>powertools-parameters</artifactId>
-            <version>{{ powertools.version }}</version>
+             <!-- TODO! Provide the parameters module you want to use here -->
+             <artifactId>powertools-parameters-secrets</artifactId>
```
jeromevdl, line 39:
```suggestion
             <artifactId>powertools-parameters-secrets</artifactId>
             <artifactId>powertools-parameters-ssm</artifactId>
             <artifactId>powertools-parameters-dynamodb</artifactId>
             <artifactId>powertools-parameters-appconfig</artifactId>
```

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
 Secrets Manager | `SecretsProvider.get(String)` `SecretsProvider.get(String, Class)`   | `secretsmanager:GetSecretValue`
 DynamoDB | `DynamoDBProvider.get(String)` `DynamoDBProvider.getMultiple(string)` | `dynamodb:GetItem` `dynamoDB:Query`
 
-## SSM Parameter Store
+## Retrieving Parameters
```
jeromevdl, line 178: regarding the permissions above, I think you don't have everything (AppConfig, and maybe others? https://docs.powertools.aws.dev/lambda/python/latest/utilities/parameters/#iam-permissions)

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
...
-
-    public class AppWithSSM implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
-        // Get an instance of the SSM Provider
-        SSMProvider ssmProvider = ParamManager.getSsmProvider();
-    
-        // Retrieve a single parameter
-        String value = ssmProvider.get("/my/parameter");
+    ```java hl_lines="9 10"
+    import com.amazonaws.services.lambda.runtime.RequestHandler;
+    import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
```
jeromevdl, line 189: for all the snippets, can we simplify, by removing these APIGW events/responses, using a simple `RequestHandler<String, String>` and adding the handler method and end of the class `}`

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
...
+    ```java hl_lines="15-19 23-28"
+    import static software.amazon.lambda.powertools.parameters.transform.Transformer.base64;
+    
+    import com.amazonaws.services.lambda.runtime.Context;
+    import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
+    import software.amazon.lambda.powertools.parameters.secrets.SecretsProvider;
+    
+    import com.amazonaws.services.lambda.runtime.RequestHandler;
+    import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+    import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
```
jeromevdl, line 259: same here, I would simplify by using Strings

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
...
+                .withClient(SsmClient.builder().build())
+                .build();
     
+        public APIGatewayProxyResponseEvent handleRequest(APIGatewayProxyRequestEvent input, Context context) {
+            // Retrieve a single secret
+            String value = ssmProvider
+                    // Transform parameter from base64
+                    .withTransformation(base64)
+                    // By default values are cached for 5 seconds, specify 10 seconds instead.
+                    .withMaxAge(10, ChronoUnit.SECONDS)
```
jeromevdl, line 316: idem, later in the advanced config

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
...
+                .withClient(AppConfigDataClient.builder().build())
+                .build();
+    
+        public APIGatewayProxyResponseEvent handleRequest(APIGatewayProxyRequestEvent input, Context context) {
+            // Retrieve a single secret
+            String value = appConfigProvider
+                    // Transform parameter from base64
+                    .withTransformation(base64)
+                    // By default values are cached for 5 seconds, specify 10 seconds instead.
+                    .withMaxAge(10, ChronoUnit.SECONDS)
```
jeromevdl, line 394: same

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
...
+        }
+    }    
     ```
 
-=== "Provider with age for each param"
-
-    ```java hl_lines="8"
-    import software.amazon.lambda.powertools.parameters.SecretsProvider;
-    import software.amazon.lambda.powertools.parameters.ParamManager;
+## Advanced configuration
```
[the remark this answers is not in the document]
jeromevdl: missing a part about writing a custom Provider

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
...
-    }
+    // Annotation-style injection from AppConfig
+    @AppConfigParam(application = "my-app", environment = "my-env", key = "myKey")
+    String appConfigParam;
     ```
 
-## Secrets Manager
 
-For secrets stored in Secrets Manager, use `getSecretsProvider`.
+### Using the `ParamProvider` classes
```
jeromevdl, line 246: missing a part about getting multiple values

docs/utilities/parameters.md
jeromevdl, line 1: See python documentation https://docs.powertools.aws.dev/lambda/python/latest/utilities/parameters, we're missing a bunch of things in ours

jeromevdl (changes requested):
API & code almost ok, documentation to improve

docs/utilities/parameters.md
scottgerring: Added missing and made more pythonic

docs/utilities/parameters.md
scottgerring: ACK

docs/utilities/parameters.md
```diff
@@ -159,260 +175,240 @@ SSM Parameter Store | `SSMProvider.getMultiple(String)`
...
+                .withClient(SecretsManagerClient.builder().build())
+                .build();
     
+        public APIGatewayProxyResponseEvent handleRequest(APIGatewayProxyRequestEvent input, Context context) {
+            // Retrieve a single secret
+            String value = secretsProvider
+                    // Transform parameter from base64
+                    .withTransformation(base64)
+                    // By default values are cached for 5 seconds, specify 10 seconds instead.
+                    .withMaxAge(10, ChronoUnit.SECONDS)
```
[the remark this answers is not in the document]
jeromevdl: I think it must be progressive, and give the mandatory stuff, and then the options, and finally going further. We can't put everything in the first example we show.

powertools-parameters/powertools-parameters-tests/pom.xml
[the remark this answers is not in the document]
scottgerring: An existing DDB test in this style didn't exist, and i've just refactored

docs/utilities/parameters.md
scottgerring: ACK; it's super redundant with the paragraph above anyway

docs/utilities/parameters.md
scottgerring: Fixed!

powertools-parameters/powertools-parameters-tests/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
scottgerring: Rejigged to restore the CacheManager::setDefaultTimeout bits where they used to be

docs/utilities/parameters.md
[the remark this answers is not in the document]
scottgerring: What puts me off is having all the "Java templateness" repeated - all the imports, RequestHandler, etc. It's just more code to maintain, and it'll be a pain when we pull it out into actual compiled code that we lint and then we have 20 variants.

But I take your point about not going straight to "extreme customization" mode. 

How about, in this block, we have two examples per provider:

1/ Extremely minimal
2/ All the config options

Something like this:

```java
     // Retrieve a single secret
     string simpleValue = secretsProvider.get("/my/secret");

     // Retrieve a single secret - with all the tuneables 
            String  lessSimpleValue = secretsProvider
                    // Transform parameter from base64
                    .withTransformation(base64)
                    // By default values are cached for 5 seconds, specify 10 seconds instead.
                    .withMaxAge(10, ChronoUnit.SECONDS)
                    .get("/my/secret");

```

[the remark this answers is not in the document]
jeromevdl: I'm sorry, I don't follow you on this. One chapter on transformation, one chapter on caching.

Don't put the imports (or eventually in the first example)

docs/utilities/parameters.md
jeromevdl: I don't find it clearer, just smaller but tons of things are missing so what's the point ?

docs/utilities/parameters.md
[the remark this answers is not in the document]
jeromevdl: If you want to do that, you must have the code somewhere in Java files, that you import somehow in the MD.... 
I did that for a CDK library project, using jsii: https://jeromevdl.medium.com/document-your-aws-cdk-constructs-like-a-pro-696fc40ed39f
But for java, mkdocs, without jsii, ... I don't know how you can do that. What would be great also, is to be able to show only a part of the "example". Java is so verbose and we don't really care about imports, we probably don't want to have 100 lines of codes for all small snippets in our doc.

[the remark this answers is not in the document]
jeromevdl: https://facelessuser.github.io/pymdown-extensions/extensions/snippets/ ==> but that's probably another PR...

docs/utilities/parameters.md
```diff
@@ -152,63 +179,80 @@ Depending on your version of Java (either Java 1.8 or 11+), the configuration sl
...
+        public string handleRequest(String request, Context context) {
+            // Because this is a secret, we probably don't want to return it! Return something indicating
+            // we could access it instead.
+            return "Retrieved a secret, and sensibly refusing to return it!";
+        }
     }
     ```
 
-=== "SSMProvider with a custom client"
+=== "Secrets Manager"
```
jeromevdl, line 225:
```suggestion
=== "Secrets Manager: SecretsProvider"
```

docs/utilities/parameters.md
```diff
@@ -217,9 +261,59 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+        @SSMParam(key = "/powertools-java/param")
+        String ssmParam;
+
+        public string handleRequest(String request, Context context) {
+            return ssmParam; // Request handler simply returns our configuration value
+        }
+    }
+    ```
+
+=== "SSM"
```
jeromevdl, line 284:
```suggestion
=== "SSM Parameter Store: SSMProvider"
```

docs/utilities/parameters.md
```diff
@@ -217,9 +261,59 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+                    // We might instead want to retrieve multiple secrets at once, returning a Map of key/value pairs
+                    // .getMultiple("/my/secret/path");
+
+            // Return the result
+            return value;
+        }
+    }
+    ```
+
+=== "SSM: Additional Options"
```
jeromevdl, line 316:
```suggestion
=== "SSM Parameter Store: Additional Options"
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,181 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
-        String value = secretsProvider.get("/my/secret");
-    
+        public string handleRequest(String request, Context context) {
+            return ddbParam;  // Request handler simply returns our configuration value
+        }
     }
     ```
 
-=== "SecretsProvider with a custom client"
+=== "DynamoDB"
```
jeromevdl, line 356:
```suggestion
=== "DynamoDB: DynamoDbProvider"
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,181 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
-To get parameters stored in AppConfig, use `getAppConfigProvider`, providing the application and environment
-name to retrieve configuration from. As with the other providers, an overloaded method allows you to retrieve
-an `AppConfigProvider` providing a client if you need to configure it yourself.
-
-=== "AppConfigProvider"
-
-    ```java hl_lines="6 9"
-    import software.amazon.lambda.powertools.parameters.AppConfigProvider;
-    import software.amazon.lambda.powertools.parameters.ParamManager;
+=== "AppConfig"
```
jeromevdl, line 409:
```suggestion
=== "AppConfig: AppConfigProvider"
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,181 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+            return value;
+        }
+    }    
     ```
 
-
 ## Advanced configuration
 
 ### Caching
+Each provider uses the CacheManager to cache parameter values. When a value is retrieved using from the provider, a
```
jeromevdl, line 442:
```suggestion
Each provider uses the `CacheManager` to cache parameter values. When a value is retrieved using from the provider, a 
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,181 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
-
 ## Advanced configuration
 
 ### Caching
+Each provider uses the CacheManager to cache parameter values. When a value is retrieved using from the provider, a 
+custom cache duration can be provided using `withMaxAge(duration, unit)`. 
 
-By default, all parameters and their corresponding values are cached for 5 seconds.
+If this is not specified, the default value set on the CacheManager itself will be used. This default can be customized
+by calling `setDefaultExpirationTime(duration, unit)` on the CacheManager.
```
jeromevdl, line 446:
```suggestion
by calling `setDefaultExpirationTime(duration, unit)` on the `CacheManager`.
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,181 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
-
-        String value = secretsProvider.get("/my/secret");
+    ```java hl_lines="14 15"
+    import java.time.Duration;
+    import software.amazon.lambda.powertools.parameters.appconfig.AppConfigProvider;
+    import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+    
+    public class CustomizeCacheTimeout {
+        
+        public void CustomizeCache() {
```
jeromevdl, line 457: do we need that ?

docs/utilities/parameters.md
```diff
@@ -238,183 +332,181 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+                    .withClient(AppConfigDataClient.builder().build())
+                    .build();
     
+            // Will use the default specified above - 10 seconds 
+            String myParam1 = paramProvider.get("myParam1");
+
+            // Will override the default above 
+            String myParam2 = paramProvider
+                .withMaxAge(20, ChronoUnit.SECONDS)
+                .get("myParam2");
```
jeromevdl, line 496: it would be enough here...

docs/utilities/parameters.md
```diff
@@ -431,14 +523,18 @@ Base64 and JSON transformations are provided. For more complex transformation, y
                         .get("/my/parameter/json", MyObj.class);
     ```
 
-## Write your own Transformer
+### Decrypt Values
+
```
jeromevdl, line 527: it's empty here

docs/utilities/parameters.md
```diff
@@ -431,14 +523,18 @@ Base64 and JSON transformations are provided. For more complex transformation, y
...
+
+
+### Create your own Transformer
```
jeromevdl, line 530:
```suggestion
#### Create your own Transformer
```

docs/utilities/parameters.md
jeromevdl, line 527:
```suggestion
```

jeromevdl: move it elsewhere...

docs/utilities/parameters.md
```diff
@@ -475,167 +571,139 @@ To simplify the use of the library, you can chain all method calls before a get.
...
+    import java.util.Map;
+    import software.amazon.lambda.powertools.parameters.BaseProvider;
+    import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+    import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
+    
+    /**
+    Our custom parameter provider itself. This does the heavy lifting of retrieving
+    parameters from whatever our underlying parameter store might be.
+    */
+    public class CustomProvider extends BaseProvider {
```
jeromevdl, line 590:
```suggestion
    /**
     *Our custom parameter provider itself. This does the heavy lifting of retrieving
     * parameters from whatever our underlying parameter store might be.
    **/
    public class CustomProvider extends BaseProvider {
```

docs/utilities/parameters.md
```diff
@@ -475,167 +571,139 @@ To simplify the use of the library, you can chain all method calls before a get.
...
+    import java.lang.annotation.Retention;
+    import java.lang.annotation.RetentionPolicy;
+    import java.lang.annotation.Target;
+    import software.amazon.lambda.powertools.parameters.transform.Transformer;
+    
+    /**
+    Aspect to inject a parameter from our custom provider. Note that if you
+    want to implement a provider _without_ an Aspect and field injection, you can
+    skip implementing both this and the {@link CustomProviderAspect} class.
+    */
```
jeromevdl, line 671:
```suggestion
    /**
     * Aspect to inject a parameter from our custom provider. Note that if you
     * want to implement a provider _without_ an Aspect and field injection, you can
     * skip implementing both this and the {@link CustomProviderAspect} class.
    **/
```

docs/utilities/parameters.md
```diff
@@ -475,167 +571,139 @@ To simplify the use of the library, you can chain all method calls before a get.
...
-                 </executions>
-            </plugin>
-            ...
-        </plugins>
-    </build>
-    ```
+    /**
+    Aspect to inject a parameter from our custom provider where the {@link CustomProviderParam}
+    annotation is used.
+    */
```
jeromevdl, line 690:
```suggestion
    /**
     * Aspect to inject a parameter from our custom provider where the {@link CustomProviderParam}
     * annotation is used.
    **/
```

docs/utilities/parameters.md
```diff
@@ -475,167 +571,139 @@ To simplify the use of the library, you can chain all method calls before a get.
...
-    repositories {
-        mavenCentral()
-    }
+        @Pointcut("get(* *) && @annotation(ddbConfigParam)")
+        public void getParam(CustomProviderParam customConfigParam) {
+        }
+    
+        @Around("getParam(customConfigParam)")
+        public Object injectParam(final ProceedingJoinPoint joinPoint, final CustomProviderParam customConfigParam) {
+            System.out.println("GET IT");
```
jeromevdl, line 700: man !

powertools-parameters/powertools-parameters-tests/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
```diff
@@ -321,8 +323,9 @@ public void getObject_customDefaultTTLAndTTL_cached_shouldGetFromCache() {
     public void getObject_customDefaultTTLAndTTL_expired_shouldGetValue() {
         provider.setValue("{\"foo\":\"Foo\", \"bar\":42, \"baz\":123456789}");
 
-        provider.defaultMaxAge(2, ChronoUnit.MINUTES)
-                .withMaxAge(5, SECONDS)
+        provider.cacheManager.setDefaultExpirationTime(Duration.ofMinutes(2));
+
+        provider.withMaxAge(5, SECONDS)
```
jeromevdl, line 328: this one is actually good, but most of the others arent

docs/utilities/parameters.md
```diff
@@ -431,14 +523,18 @@ Base64 and JSON transformations are provided. For more complex transformation, y
...
+
+### Create your own Transformer
 
 You can write your own transformer, by implementing the `Transformer` interface and the `applyTransformation()` method.
 For example, if you wish to deserialize XML into an object.
 
 === "XmlTransformer.java"
 
-    ```java hl_lines="1"
+    ```java
```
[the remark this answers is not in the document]
scottgerring: Yep! Catch

docs/utilities/parameters.md
scottgerring: 🤣

powertools-parameters/powertools-parameters-appconfig/src/main/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderBuilder.java
scottgerring: Nope!

powertools-parameters/powertools-parameters-appconfig/src/test/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderTest.java
```diff
@@ -189,27 +188,15 @@ public void multipleKeysRetrievalWorks() {
     public void getMultipleValuesThrowsException() {
 
         // Act & Assert
-        assertThatRuntimeException().isThrownBy(() -> provider.getMultipleValues("path"))
+        Assertions.assertThatRuntimeException().isThrownBy(() -> provider.getMultipleValues("path"))
```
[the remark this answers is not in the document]
scottgerring: 🤷

powertools-parameters/powertools-parameters-dynamodb/pom.xml
scottgerring: Where have I heard this before

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbProvider.java
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @param path Parameter store path
+     * @return All values matching the given path, and an empty map if none do. Throws if any records exist that don't match the schema.
+     */
+    @Override
+    protected Map<String, String> getMultipleValues(String path) {
+
+        QueryResponse resp = client.query(QueryRequest.builder()
+                .tableName(tableName)
+                .keyConditionExpression("id = :v_id")
```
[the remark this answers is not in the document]
scottgerring: They'd have different sort keys

powertools-parameters/powertools-parameters-tests/src/test/java/software/amazon/lambda/powertools/parameters/BaseProviderTest.java
```diff
@@ -173,9 +172,10 @@ public void get_customDefaultTTLAndTTL_cached_shouldGetFromCache() {
 
     @Test
     public void get_customDefaultTTLAndTTL_expired_shouldGetValue() {
-        provider.defaultMaxAge(2, ChronoUnit.MINUTES)
-                .withMaxAge(5, SECONDS)
-                .get("bariton");
+
+        provider.cacheManager.setDefaultExpirationTime(Duration.ofSeconds(5));
```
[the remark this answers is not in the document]
scottgerring: 🤦

examples/powertools-examples-parameters/src/main/java/org/demo/parameters/ParametersFunction.java
scottgerring: Seems reasonable. Implementing!

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/BaseProvider.java
scottgerring: It is used

docs/utilities/parameters.md
scottgerring: Dropped as the single complete example covers this

docs/utilities/parameters.md
```diff
@@ -152,63 +179,80 @@ Depending on your version of Java (either Java 1.8 or 11+), the configuration sl
...
+    public class ParametersFunction implements RequestHandler<String, String> {
+
+        // Annotation-style injection from secrets manager
+        @SecretsParam(key = "/powertools-java/userpwd")
+        String secretParam;
+
+        public string handleRequest(String request, Context context) {
+            // Because this is a secret, we probably don't want to return it! Return something indicating
+            // we could access it instead.
+            return "Retrieved a secret, and sensibly refusing to return it!";
```
jeromevdl, line 220: can we avoid this kind of long, somehow useless, comment?
```suggestion
            // ... use the secretParam 
            return "something";
```

docs/utilities/parameters.md
```diff
@@ -152,63 +179,80 @@ Depending on your version of Java (either Java 1.8 or 11+), the configuration sl
...
-        // Retrieve multiple parameters from a path prefix
-        // This returns a Map with the parameter name as key
-        Map<String, String> values = ssmProvider.getMultiple("/my/path/prefix");
+        public String handleRequest(String input, Context context) {
+            // Retrieve a single secret
+            String value = secretsProvider.get("/my/secret");
     
+            // Because this is a secret, we probably don't want to return it! Return something indicating
+            // we could access it instead.
+            return "Retrieved a secret, and sensibly refusing to return it!";
```
jeromevdl, line 250: same here

docs/utilities/parameters.md
```diff
@@ -217,9 +261,59 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+    
+        // Get an instance of the SSMProvider. We can provide a custom client here if we want,
+        // for instance to use a particular region.
+        SSMProvider ssmProvider = SSMProvider
+                .builder()
+                .withClient(SsmClient.builder().build())
+                .build();
+    
+        public String handleRequest(String input, Context context) {
+            // Retrieve a single secret
```
jeromevdl, line 304: not a secret

docs/utilities/parameters.md
```diff
@@ -217,9 +261,59 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+            String value = ssmProvider
+                    .get("/my/secret");
+                    // We might instead want to retrieve multiple secrets at once, returning a Map of key/value pairs
```
jeromevdl, line 307: not secrets
```suggestion
                    // We might instead want to retrieve multiple parameters at once, returning a Map of key/value pairs
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,159 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
 
-    public class AppWithSecrets implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
-        SecretsManagerClient client = SecretsManagerClient.builder().region(Region.EU_CENTRAL_1).build();
-        // Get an instance of the Secrets Provider
-        SecretsProvider secretsProvider = ParamManager.getSecretsProvider(client);
+    public class RequestHandlerWithParams implements RequestHandler<String, String> {
     
-        // Retrieve a single secret
-        String value = secretsProvider.get("/my/secret");
+        // Get an instance of the SecretsProvider. We can provide a custom client here if we want,
```
jeromevdl, line 368:
```suggestion
        // Get an instance of the DynamoDbProvider. We can provide a custom client here if we want,
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,159 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+        // for instance to use a particular region.
+        DynamoDbProvider ddbProvider = DynamoDbProvider
+                .builder()
+                .withClient(DynamoDbClient.builder().build())
+                .build();
     
+        public String handleRequest(String input, Context context) {
+            // Retrieve a single secret
```
jeromevdl, line 376:
```suggestion
            // Retrieve a single param
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,159 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
-                .httpClientBuilder(UrlConnectionHttpClient.builder())
-                .region(Region.EU_CENTRAL_2)
-                .build();
+    import software.amazon.lambda.powertools.parameters.appconfig.AppConfigProvider;
 
-        // Get an instance of the DynamoDbProvider
-        AppConfigProvider appConfigProvider = ParamManager.getAppConfigProvider(appConfigDataClient, "my-environment", "my-app");
+    public class RequestHandlerWithParams implements RequestHandler<String, String> {
+    
+        // Get an instance of the SecretsProvider. We can provide a custom client here if we want,
```
jeromevdl, line 421:
```suggestion
        // Get an instance of the AppConfigProvider. We can provide a custom client here if we want,
```

docs/utilities/parameters.md
```diff
@@ -238,183 +332,159 @@ The AWS Systems Manager Parameter Store provider supports two additional argumen
...
+        AppConfigProvider appConfigProvider = AppConfigProvider
+                .builder()
+                .withClient(AppConfigDataClient.builder().build())
+                .build();
     
-        // Retrieve a single parameter
-        String value = appConfigProvider.get("my-key"); 
-    } 
+        public String handleRequest(String input, Context context) {
+            // Retrieve a single secret
```
jeromevdl, line 429:
```suggestion
            // Retrieve a single param
```

powertools-parameters/powertools-parameters-appconfig/src/test/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderTest.java
```diff
@@ -12,19 +12,20 @@
...
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
+import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
```
jeromevdl, line 28:
```suggestion
```

powertools-parameters/powertools-parameters-appconfig/src/test/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderTest.java
```diff
@@ -12,19 +12,20 @@
...
-package software.amazon.lambda.powertools.parameters;
+package software.amazon.lambda.powertools.parameters.appconfig;
 
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
 import static org.assertj.core.api.Assertions.assertThatRuntimeException;
-import static org.mockito.MockitoAnnotations.openMocks;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
+import org.assertj.core.api.Assertions;
```
jeromevdl, line 21:
```suggestion
```

powertools-parameters/powertools-parameters-appconfig/src/test/java/software/amazon/lambda/powertools/parameters/appconfig/AppConfigProviderTest.java
```diff
@@ -52,7 +53,7 @@ public class AppConfigProviderTest {
 
     @BeforeEach
     public void init() {
-        openMocks(this);
+        MockitoAnnotations.openMocks(this);
```
jeromevdl, line 56:
```suggestion
       openMocks(this);
```

powertools-parameters/powertools-parameters-dynamodb/src/main/java/software/amazon/lambda/powertools/parameters/dynamodb/DynamoDbParam.java
```diff
@@ -21,25 +21,31 @@
...
- * The library provide a provider for AWS System Manager Parameters Store ({@link SSMProvider}) and a provider
- * for AWS Secrets Manager ({@link SecretsProvider}).
- * The user can implement a custom provider by extending the abstract class {@link BaseProvider}.</p>
- *
- * <p>If the parameter value requires transformation before being assigned to the annotated field
- * users can specify a {@link Transformer}
- * </p>
+ * Usage:
+ * <pre>
+ * @SSMParam(key = "my-param", table = "my-table")
```
jeromevdl, line 29:
```suggestion
 * @DynamoDbParam(key = "my-param", table = "my-table")
```

jeromevdl (commented):
almost good

docs/utilities/parameters.md
scottgerring:
```suggestion
            // ... do something with the secretParam here
            return "something";
```

jeromevdl (approved):
LGTM

jeromevdl:
:bananadance:

PR #1409, opened 2023-08-30
chore: update v2

jeromevdl:
**Issue #, if available:**

## Description of changes:

update v2 with main

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


examples/powertools-examples-core-utilities/cdk/app/pom.xml
```diff
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>software.amazon.lambda.examples</groupId>
-    <version>1.16.1</version>
+    <version>1.17.0</version>
```
scottgerring, line 6: This is wrong (but it's also wrong in `v2`) . Should we fix it here?

scottgerring: It appears to only be this example, which follows as it was added after we branched v2

examples/powertools-examples-core-utilities/sam/pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>software.amazon.lambda.examples</groupId>
     <version>2.0.0-SNAPSHOT</version>
-    <artifactId>powertools-examples-core-utilities</artifactId>
+    <artifactId>powertools-examples-core-sam</artifactId>
```
scottgerring, line 7:
```suggestion
    <artifactId>powertools-examples-core-utilities-sam</artifactId>
```
probably? Not sure where this change has come from though

scottgerring (commented):
Two comments inline - one was already there and just noticed it because the merge touches it (version in POM)

examples/powertools-examples-core-utilities/cdk/app/pom.xml
jeromevdl: it does not work if I put 2.0.0-SNAPSHOT, because it cannot find the installed version (2.0.0-SNAPSHOT) of the libraries. I didn't dig into why, but probably because it is built in docker and there's no v2 in central...

jeromevdl: let's see that apart, so that we finish the merge

scottgerring (approved):

PR #1451, opened 2023-09-28
test: remove x-ray from example tests and add LAMBDA_TASK_ROOT

AlexeySoshin:
Add a segment before the test starts and end it once it finishes, same as with Maven examples

**Issue #, if available:**

## Description of changes:

Other examples have setup/teardown functions
https://github.com/aws-powertools/powertools-lambda-java/blob/ef11a64df17ee1cbf736aacbd308c9f191d21a8c/examples/powertools-examples-core/sam/src/test/java/helloworld/AppTest.java#L30

**Checklist**
* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


jeromevdl:
To be frank, I wanted to remove tests in samples, but maybe we should have better tests instead.

The tracing stuff is typically the kind of boring things. Idempotency with dynamodb also (we had a question about that), ...

And it's funny because we have samples containing code that we don't recommend: https://docs.powertools.aws.dev/lambda/java/core/tracing/#configure-test-cases-not-recommended

scottgerring:
I thought we should just merge this for consistency, but now I am less sure 🤣 I think it is 1/ nice that we have a test that exercises the handler but 2/ think the x-ray stuff is a bit overwrought. A third option would be keeping the tests and removing the x-ray tracing from all of them, which lines us up with our documented best practices.

@jeromevdl ?

jeromevdl:
then we must add the property in the pom.xml: https://docs.powertools.aws.dev/lambda/java/core/tracing/#configure-environment-variable-on-project-level-recommended

AlexeySoshin:
@jeromevdl I have added this to the Gradle example for now:
https://github.com/aws-powertools/powertools-lambda-java/pull/1451/files#diff-80b08d79d09d03fb438f451205323a1e77e99ba8ffe4842545d235b5b2790989R37

scottgerring:
@AlexeySoshin I think you should remove x-ray from all the example tests (right @jeromevdl ? )

examples/powertools-examples-core/gradle/build.gradle
```diff
@@ -11,10 +11,6 @@ wrapper {
 compileJava {
     sourceCompatibility = "1.8"
     targetCompatibility = "1.8"
-
-    ajc {
```
AlexeySoshin, line 15: @scottgerring FYI, I think this is not needed as you already use `aspect`

scottgerring: If it works without it let's get rid of it!

jeromevdl:
we removed all the tests in the example, I think we can close it.

Issue #1298, opened 2023-07-19
Bug: validation module should return 400 on error


scottgerring:
When you use the `@Validation` annotation on a `RequestHandler` and the provided message does not validate, the module will throw a `ValidationException`, resulting in a 500 back to the client. 

Changing this behaviour is a breaking change.

**What were you trying to accomplish?**
I would expect that validation errors should return a 400 and a description of why the validation failed. Alternatively if the annotation is applied to something other than a `RequestHandler`, throwing an exception for the user's code to catch seems reasonable. 

## Current Behavior
ValidationException / HTTP 500 returned to user. 

## Steps to Reproduce (for bugs)
Check out the [validation example](https://github.com/aws-powertools/powertools-lambda-java/tree/main/examples/powertools-examples-validation) as it exhibits this behaviour.

jeromevdl:
Not that simple as we perform some validation on many different events, not only API GW... 

400 is only applicable to API GW events and responses (`APIGatewayProxyRequestEvent`, `APIGatewayV2HTTPEvent`, `APIGatewayProxyResponseEvent` and `APIGatewayV2HTTPResponse`), otherwise an exception is still valid.

scottgerring:
Good point. Should we add some return-type specific behavior?

 The alternative seems to be saying we aren't supporting the annotation directly on the handlers function itself, but only on "other methods"; then in the case of a HTTP API integration, the user's handler wraps this "other method", catches the exception, and maps back to a 400.

 I feel that validation on a HTTP API is "common path" and should work without this level of glue. What do you reckon?

jeromevdl:
Either we document how to **not** use the annotation and use the `ValidationUtils` in case of API and catch the validation exception to return 400. This is what python did ([doc](https://docs.powertools.aws.dev/lambda/python/latest/utilities/validation/#validate-function))

Either we implement this in the library : in case of one of the events above, we change the http code in the response to 400 and update the body with an error message... Probably a better experience for users but will be opinionated (message format: json? plain txt?).

scottgerring:
Let's tie this into the broader module-review issue for V2 - #1469. Oner person can knock both of these off at once.

skal111:
Happy to take that one

scottgerring:
This will be released with v2

PR #1489, opened 2023-10-24
feat(v2): Validation failures return 400s 

skal111:
api gateway events validation will be catched and returned as a 400 error

#1298

## Description of changes:

ValidationException for events of type APIGateway are catched and transformed into a valid response (status 400).

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/ValidationE2ET.java
```diff
@@ -0,0 +1,99 @@
...
+		if (infrastructure != null) {
+			infrastructure.destroy();
+		}
+	}
+
+	@ParameterizedTest
+	@Event(value = "/validation/valid_api_gw_in_out_event.json", type = APIGatewayProxyRequestEvent.class)
+	void test_validInboundApiGWEvent(APIGatewayProxyRequestEvent validEvent) throws IOException {
+		// WHEN
+		InvocationResult invocationResult = invokeFunction(functionName, objectMapper.writeValueAsString(validEvent));
```
jeromevdl, line 65: you actually deserialize with the `@Event` annotation and reserialize here, you can probably just load the file as string to avoid this double step...

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/ValidationE2ET.java
```diff
@@ -0,0 +1,99 @@
...
+	void test_invalidOutboundApiGWEvent(APIGatewayProxyRequestEvent validEvent) throws IOException {
+		// WHEN
+		InvocationResult invocationResult = invokeFunction(functionName, objectMapper.writeValueAsString(validEvent));
+
+		// THEN
+		// invocation should fail outbound validation and return 400
+		JsonNode validJsonNode = objectMapper.readTree(invocationResult.getResult());
+		assertThat(validJsonNode.get("statusCode").asInt()).isEqualTo(400);
+		assertThat(validJsonNode.get("body").asText()).contains("$.price: must have an exclusive maximum value of 1000");
+	}
```
jeromevdl, line 98: could we add another test with something else than API GW, to make sure there is an exception thrown?

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/ValidationE2ET.java
skal111: I can replace loading the event with an InputStream. The following works as well.

`@Test
	void test_validInboundApiGWEvent() throws IOException {
		InputStream inputStream = this.getClass().getResourceAsStream("/validation/valid_api_gw_in_out_event.json");
		String validEvent = IOUtils.toString(inputStream, StandardCharsets.UTF_8);

		// WHEN
		InvocationResult invocationResult = invokeFunction(functionName, validEvent);
`

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/ValidationE2ET.java
skal111: The Function handler needs to be typed for the @Validation to works. Which means I need to add a second test function with something else than APIGW. I'm just not sure how to do it (in a compliant way) with the underlying Infrastructure builder that takes a single pathToFunction argument

docs/utilities/validation.md
```diff
@@ -156,7 +156,9 @@ We support JSON schema version 4, 6, 7 and 201909 (from [jmespath-jackson librar
 
 `@Validation` annotation is used to validate either inbound events or functions' response.
 
-It will fail fast with `ValidationException` if an event or response doesn't conform with given JSON Schema.
+It will fail fast if an event or response doesn't conform with given JSON Schema. For most type of events a `ValidationException` will be thrown. But for
```
scottgerring, line 159:
```suggestion
It will fail fast if an event or response doesn't conform with given JSON Schema. For most type of events a `ValidationException` will be thrown. 
```

docs/utilities/validation.md
```diff
@@ -156,7 +156,9 @@ We support JSON schema version 4, 6, 7 and 201909 (from [jmespath-jackson librar
 
 `@Validation` annotation is used to validate either inbound events or functions' response.
 
-It will fail fast with `ValidationException` if an event or response doesn't conform with given JSON Schema.
+It will fail fast if an event or response doesn't conform with given JSON Schema. For most type of events a `ValidationException` will be thrown. But for 
+specific `APIGatewayProxyRequestEvent` and `APIGatewayV2HTTPEvent`, instead, the `@Validation` annotation will build and return a custom
```
scottgerring, line 160:
```suggestion
For API gateway events associated with REST APIs and HTTP APIs -  `APIGatewayProxyRequestEvent` and `APIGatewayV2HTTPEvent` - the `@Validation` annotation will build and return a custom 400 / "Bad Request" response, with a body containing the validation errors. This saves you from having to catch the validation exception and map it back to a meaningful user error yourself.
```

scottgerring: Trying to make it explicit what is happening and why!

docs/utilities/validation.md
```diff
@@ -156,7 +156,9 @@ We support JSON schema version 4, 6, 7 and 201909 (from [jmespath-jackson librar
...
+400 (BAD_REQUEST) response. Its body will contain the validation error(s) message(s).
```
scottgerring, line 161:
```suggestion
```

powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/internal/ValidationAspect.java
```diff
@@ -140,33 +146,88 @@ record -> validate(decode(record.getData()), inboundJsonSchema)));
...
+        else {
+        	result = pjp.proceed(proceedArgs);
+        	
+        	if (validationNeeded && !validation.outboundSchema().isEmpty()) {
+        		JsonSchema outboundJsonSchema = getJsonSchema(validation.outboundSchema(), true);
+        		
+        		Object overridenResponse = null;
+        		if (result instanceof APIGatewayProxyResponseEvent) {
+        			APIGatewayProxyResponseEvent response = (APIGatewayProxyResponseEvent) result;
+        			overridenResponse = validateAPIGatewayProxyBody(response.getBody(), outboundJsonSchema);
```
jeromevdl, line 163: Question: if the response already contains some information that needs to be returned, should we actually get rid of that? For example if they add some headers, we probably want to pass them. Maybe you should pass the response as a parameter and just override the body/status code, but keep the rest.

powertools-validation/src/main/java/software/amazon/lambda/powertools/validation/internal/ValidationAspect.java
```diff
@@ -140,33 +146,88 @@ record -> validate(decode(record.getData()), inboundJsonSchema)));
...
+        		} else if (result instanceof APIGatewayV2HTTPResponse) {
+        			APIGatewayV2HTTPResponse response = (APIGatewayV2HTTPResponse) result;
+        			overridenResponse = validateAPIGatewayV2HTTPBody(response.getBody(), outboundJsonSchema);
```
jeromevdl, line 166: same here

powertools-validation/src/test/java/software/amazon/lambda/powertools/validation/internal/HandledResponseEventsArgumentsProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayV2HTTPResponse;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayV2WebSocketResponse;
+import com.amazonaws.services.lambda.runtime.events.ApplicationLoadBalancerResponseEvent;
+
+public class HandledResponseEventsArgumentsProvider implements ArgumentsProvider {
```
jeromevdl, line 28: can you add a comment here? what is it for ?

jeromevdl (commented):
See above comments.

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/ValidationE2ET.java
jeromevdl: It will require a new function and a new E2ET.

scottgerring:
> Additional point that we completely forgot: it would be great to make the `@Validation` annotation compatible with the batch module (SQS, Kinesis), i.e. if one message is wrong do not fail the whole batch but add the message to the partial failure list.

This is a great idea - I think PT should aspire to "magically do what you'd expect" when you combine modules. We'd need to pay special attention to the coupling between the modules - explicitly maven coupling between batch<->validation is IMHO a non-starter. Perhaps we could pull validation exceptions into their own package, or something 🤷  - i've not thought about this in depth, but I think how this coupling would work will be decisive here.

@jeromevdl pragmatically we could also treat this integration in a subsequent PR. WDYT?

jeromevdl:
I agree we could do that in another PR. 

Regarding the coupling, **there should not be any dependency between the 2 modules**. And I think there won't be any.

1. We can keep the validation code as today (almost) but rather than throwing an exception, do just like @skal111 did for API GW, ie. creating an `SQSBatchResponse` / `StreamsEventResponse` and put the messages in error in the partial error. 

2. Then the handler is called with all the remaining messages (or directly return if no message is valid), and here the batch module will work. 

3. In the end the validation module completes the `SQSBatchResponse` / `StreamsEventResponse` returned by the handler (batch module) with the validation failure from #1. 

Finger in the nose ?!

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/ValidationE2ET.java
skal111: Added an E2E test for ALB events which demonstrate that in that case an exception is thrown

powertools-e2e-tests/handlers/validation-alb-event/src/main/java/software/amazon/lambda/powertools/e2e/Function.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.ApplicationLoadBalancerRequestEvent;
+import com.amazonaws.services.lambda.runtime.events.ApplicationLoadBalancerResponseEvent;
+import com.amazonaws.services.lambda.runtime.events.SQSBatchResponse;
+
+import software.amazon.lambda.powertools.validation.Validation;
+public class Function implements RequestHandler<ApplicationLoadBalancerRequestEvent, ApplicationLoadBalancerResponseEvent> {
+//	@Validation(inboundSchema = "classpath:/validation/inbound_schema.json", outboundSchema = "classpath:/validation/outbound_schema.json")
```
jeromevdl, line 25: can we cleanup this comment?

powertools-e2e-tests/handlers/validation-alb-event/src/main/java/software/amazon/lambda/powertools/e2e/Function.java
```diff
@@ -0,0 +1,40 @@
...
+//	public String handleRequest(SQSEvent input, Context context) {
+//		return "OK";
+//	}
+	
+	
+	@Validation(inboundSchema = "classpath:/validation/inbound_schema.json", outboundSchema = "classpath:/validation/outbound_schema.json")
+//	@Validation(inboundSchema = "classpath:/validation/inbound_schema.json")
```
jeromevdl, line 32: can we cleanup this comment too?

jeromevdl (commented):
last couple of things + rebase v2 and we should be good

jeromevdl (approved):
LGTM, thank you @skal111

Issue #1499, opened 2023-10-31
Trace Annotations not getting added if Annotation Key has a space in it


thetuxwhocodes:
**What were you trying to accomplish?**

I noticed that Trace Annotations not getting added if Annotation Key has a space in it

The following code works
            TracingUtils.putAnnotation("TraceName", "MyTrace");
The following code doesn't work
             TracingUtils.putAnnotation("Trace Name", "MyTrace");

## Expected Behavior

Annotation should be visible in Trace details page

Annotation should be visible in Trace details page irrespective of how the key is created. 

## Current Behavior

Annotation is totally missing if key has a space in it

## Possible Solution

Did not dig deep on the code

## Steps to Reproduce (for bugs)

1. Try with following code.
2.            TracingUtils.putAnnotation("TraceName", "MyTrace");
3.
4.

## Environment

* **Powertools for AWS Lambda (Java) version used**: 1.17.0
* **Packaging format (Layers, Maven/Gradle)**: Maven
* **AWS Lambda function runtime:** java11
* **Debugging logs**

> [How to enable debug mode](https://docs.powertools.aws.dev/lambda-java/#debug-mode)**

```text
# paste logs here
```

thetuxwhocodes:
Steps to Reproduce (for bugs)

    Try with following code.

           TracingUtils.putAnnotation("Trace Name", "MyTrace");

scottgerring:
Hey @thetuxwhocodes , thanks for raising the issue!

It [looks like annotation keys have to be alphanumeric](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-annotations) without spaces for X-Ray. I'm not sure what we should do here - the underlying x-ray library isn't validating this - should we? What do you think?

msailes:
I feel like the X-Ray library should validate this. Powertools also has the chance to help customers avoid this problem.

thetuxwhocodes:
The confusion arises when space is allowed for metadata key and not for annotations. If currently space cannot be supported, I would suggest updating the powertools documentation to make it clear.

scottgerring:
@msailes ACK your point about improving user XP. I feel like it belongs primarily in x-ray and will raise a ticket over there also. 

If PT were to validate it, the result would likely be a runtime exception. At least then you'd have a some immediate indication of the problem, rather than a silent failure and a bug hunt. What do you both think about this?

[...]

PR #1512, opened 2023-11-10
chore: Addition of Warn Message If Invalid Annotation Key While Tracing #1511

jdoherty:
[**Issue #, 1499:**](https://github.com/aws-powertools/powertools-lambda-java/issues/1499)

## Description of changes:

Change to emit warning log message if an invalid key is used while putting an annotation while tracing.  

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-tracing/pom.xml
+++ b/powertools-tracing/pom.xml
@@ -48,6 +48,10 @@
         </developer>
     </developers>
 
+    <properties>
+        <logback.version>1.2.11</logback.version>
+    </properties>
+
     <distributionManagement>
         <snapshotRepository>
             <id>ossrh</id>
@@ -115,6 +119,13 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>${logback.version}</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
--- a/powertools-tracing/pom.xml
+++ b/powertools-tracing/pom.xml
@@ -48,10 +48,6 @@
         </developer>
     </developers>
 
-    <properties>
-        <logback.version>1.2.11</logback.version>
-    </properties>
-
     <distributionManagement>
         <snapshotRepository>
             <id>ossrh</id>
@@ -119,13 +115,6 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
-            <version>${logback.version}</version>
-            <scope>test</scope>
-        </dependency>
-
     </dependencies>
 
     <build>
--- a/powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
+++ b/powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
@@ -39,7 +39,7 @@ public final class TracingUtils {
      * @param value the value of the annotation
      */
     public static void putAnnotation(String key, String value) {
-        if (!isValidateAnnotationKey(key)) {
+        if (!isValidAnnotationKey(key)) {
             LOG.warn("Ignoring annotation with unsupported characters in key: {}", key);
             return;
         }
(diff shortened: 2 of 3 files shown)


powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
```diff
@@ -36,6 +39,7 @@ public final class TracingUtils {
      * @param value the value of the annotation
      */
     public static void putAnnotation(String key, String value) {
+        validateAnnotationKey(key);
```
scottgerring, line 42: Should `validateAnnotationKey` returns `true`/`false`, and skip calling `putAnnotation` if it is false? 

From the issue description it seems like the behavior will be the same, but it would make it more explicit when reading through what's going on.

powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
```diff
@@ -47,10 +51,17 @@ public static void putAnnotation(String key, String value) {
      * @param value the value of the annotation
      */
     public static void putAnnotation(String key, Number value) {
+        validateAnnotationKey(key);
         AWSXRay.getCurrentSubsegmentOptional()
                 .ifPresent(segment -> segment.putAnnotation(key, value));
     }
 
+    private static void validateAnnotationKey(String key) {
```
scottgerring, line 59:
```suggestion
   /**
   Make sure that the annotation key is valid according to 
   <a href='https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-annotations'>the documentation</a>.
   
   Annotation keys that are added that are invalid are ignored by x-ray.
   **/
    private static void validateAnnotationKey(String key) {
```

powertools-tracing/pom.xml
```diff
@@ -115,6 +119,13 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
scottgerring, line 122: FYI:
in `main` (v1) we are using log4j everywhere, in `v2` we are switching to slf4j. I would say it would be nice to keep this with a log4j in memory appender or so for the test, but seeing as we would then pull it immediately out and replace it with slf4j based stuff anyway and we're in a test, it seems a bit silly 👍

scottgerring (changes requested):
Couple of small observations but otherwise looks good!

powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
jeromevdl: +1

powertools-tracing/pom.xml
jeromevdl: +1000

powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
jdoherty: I originally did this but changed it because I needed to have if branching and log statements in two methods rather than consolidating in one place.  But I've changed it back now.

powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
jdoherty: done

powertools-tracing/src/main/java/software/amazon/lambda/powertools/tracing/TracingUtils.java
```diff
@@ -47,10 +51,17 @@ public static void putAnnotation(String key, String value) {
...
+        if (!key.matches("^[a-zA-Z0-9_]+$")) {
+            LOG.warn("ignoring annotation with unsupported characters in key: {}", key);
```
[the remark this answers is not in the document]
jdoherty: done

scottgerring:
LGTM! Will merge when the build passes. Thanks @jdoherty !

powertools-tracing/pom.xml
jeromevdl: can we remove this? Please use the log4j-slf4j-impl that we have elsewhere...

jeromevdl (changes requested):
remove logback and we should be good

jeromevdl (approved):

PR #1518, opened 2023-11-17
fix: LargeMessageIdempotentE2ET Flaky 

scottgerring:
**Issue #, if available:**
n/a

## Description of changes:
The test fails intermittently. If you look at the state of the world after the test fails, the data it is looking for in the DynamoDB table is there. This indicates that the test is failing because we are not giving the Lambda enough time to complete.

Bumping out timing magic numbers is not very satisfying but I can't see another way of doing this.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


jeromevdl (approved):
flacky / not flacky, that is the question, but let's go!

PR #1532, opened 2023-11-28
docs: HelloWorldStreamFunction in examples fails with sam

jasoniharris:
**Issue #, if available:**

## Description of changes:

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


.github/workflows/pr_build.yml
```diff
@@ -79,33 +79,12 @@ jobs:
         if: ${{ matrix.java == '8' }} # Gradle example can only be built on Java 8
         working-directory: examples/powertools-examples-core/kotlin
         run: ./gradlew build
-      - name: Setup Terraform
```
scottgerring, line 82: Hey @jasoniharris does this PR  include the contents of #1533?

examples/powertools-examples-core/sam/src/main/java/helloworld/AppStream.java
```diff
@@ -17,22 +17,40 @@
...
 public class AppStream implements RequestStreamHandler {
     private static final ObjectMapper mapper = new ObjectMapper();
+    private final static Logger log = LogManager.getLogger(AppStream.class);
 
     @Override
     @Logging(logEvent = true)
     @Metrics(namespace = "ServerlessAirline", service = "payment", captureColdStart = true)
-    public void handleRequest(InputStream input, OutputStream output, Context context) throws IOException {
-        Map map = mapper.readValue(input, Map.class);
+    public void handleRequest(InputStream input, OutputStream output, Context context) {
```
scottgerring, line 44: I think it would be helpful to add some info here about why you would use `RequestStreamHandler` - e.g. to make it clear that it has nothing to do with HTTP streaming.

.github/workflows/pr_build.yml
jasoniharris: Yes - My mistake - Just pushed to revert this to current main.

examples/powertools-examples-core/sam/src/main/java/helloworld/AppStream.java
jasoniharris: Note added to explain why RequestStreamHandler is preferred here

scottgerring: I think it's not quite right - I think the use case is more typically "we need to control [de]serialization ourselves".

 I find "rather than a simple HTTP request-response model" a little confusing, because you still have exactly the same surrounding HTTP interaction as you would with a 'normal' RequestHandler, you're just choosing to read-and-write the `Body` using java streams rather than letting Lambda handle the serialization for you.

@jeromevdl I don't know if you have some super clear wording handy? The core Lambda documentation isn't that prescriptive.

jeromevdl: I agree wording is not clear or even wrong. `RequestStreamHandler` is not linked to [Lambda streaming](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/) which was announced recently. It is used when you want to use your own serialization: https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html#java-handler-interfaces. I agree the name of the interface is not super well chosen but it has nothing to do with streams...

examples/powertools-examples-core/sam/src/main/java/helloworld/AppStream.java
```diff
@@ -17,22 +17,43 @@
...
+    private final static Logger log = LogManager.getLogger(AppStream.class);
 
     @Override
     @Logging(logEvent = true)
     @Metrics(namespace = "ServerlessAirline", service = "payment", captureColdStart = true)
-    public void handleRequest(InputStream input, OutputStream output, Context context) throws IOException {
-        Map map = mapper.readValue(input, Map.class);
+    // RequestStreamHandler used from the AWS Lambda runtime for handling streaming events as it can process the input
+    // and output streams directly. Preferred when output data is represented as a stream of bytes,
+    // rather than a simple HTTP request-response model.
```
AlexeySoshin, line 46:
```suggestion
    // RequestStreamHandler can be used instead of RequestHandler for cases when you'd like to deserialize request body or serialize response body yourself, instead of allowing that to happen automatically
    // Note that you still need to return a proper JSON for API Gateway to handle 
    // See Lambda Response format for examples: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
```

scottgerring (approved):
LGTM!

Issue #1463, opened 2023-10-10
Maintenance: Clean up POMs


scottgerring:
### Summary

There is a lot of information duplicated between all of our POMs that can likely be kept in the parent POM only. We should take the v2 release as an opportunity to clean it up. Using `powertools-core` as an example, this all seems unnecessary:

https://github.com/aws-powertools/powertools-lambda-java/blob/349ea0b937a1b39b7b8926ae6985f4545c9fe33f/powertools-core/pom.xml#L34-L56

If we have a look at the `powertools-batch` [POM](https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-batch/pom.xml), we can see that it's working fine without any of this, as its all in the parent anyway.

### Why is this needed?

As we added more projects (examples, modules broken down into smaller components), the amount of repetition in the POMs themselves is becoming increasingly error-prone.

### Which area does this relate to?

Governance

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda (Java) languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

dreamorosi:
@phipag I see there's already a merged PR mentioned in the events above, can this be closed?

PR #1535, opened 2023-12-04
chore: cleanup poms and reduce warning noise

jdoherty:
**Issue #**: #1463

## Description of changes:

Removed duplicated information from child poms that was present in parent pom. 
Added version of plugin and optimised imports to reduce warning noises

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -48,6 +48,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-deploy-plugin</artifactId>
+                <version>3.1.1</version>
                 <configuration>
                     <skip>true</skip>
                 </configuration>
--- a/powertools-cloudformation/pom.xml
+++ b/powertools-cloudformation/pom.xml
@@ -32,28 +32,6 @@
         A suite of utilities for AWS Lambda Functions that makes tracing with AWS X-Ray, structured logging and creating
         custom metrics asynchronously easier.
     </description>
-    <url>https://aws.amazon.com/lambda/</url>
-    <issueManagement>
-        <system>GitHub Issues</system>
-        <url>https://github.com/aws-powertools/powertools-lambda-java/issues</url>
-    </issueManagement>
-    <scm>
-        <url>https://github.com/aws-powertools/powertools-lambda-java.git</url>
-    </scm>
-    <developers>
-        <developer>
-            <name>Powertools for AWS Lambda (Java) Team</name>
-            <organization>Amazon Web Services</organization>
-            <organizationUrl>https://aws.amazon.com/</organizationUrl>
-        </developer>
-    </developers>
-
-    <distributionManagement>
-        <snapshotRepository>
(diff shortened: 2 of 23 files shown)


examples/pom.xml
```diff
@@ -48,6 +48,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-deploy-plugin</artifactId>
+                <version>3.1.1</version>
```
scottgerring, line 51: I feel a bit funny about this one. Some context: we only include this bit so that we don't deploy the examples as part of our maven release process. That is - when these are being built inside of our own maven reactor. 

Optimally when a user copied these and used them, there'd be nothing at all in here about maven deployment. 

I suppose my question is - what warnings does the lack of version here produce? Do we see them 1/ when running within this project's overarching maven reactor, 2/ when running _separately_, that is, copied out of it, or 3/ in both cases?

powertools-cloudformation/pom.xml
```diff
@@ -32,28 +32,6 @@
         A suite of utilities for AWS Lambda Functions that makes tracing with AWS X-Ray, structured logging and creating
         custom metrics asynchronously easier.
     </description>
-    <url>https://aws.amazon.com/lambda/</url>
```
scottgerring, line 35: Great to see this get cleaned up! So much noise and duplication that makes it hard to see the woods for the trees.

powertools-e2e-tests/pom.xml
```diff
@@ -184,6 +184,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-deploy-plugin</artifactId>
+                <version>3.1.1</version>
```
scottgerring, line 187: For "internal" stuff - not the examples - I don't know if it makes sense to pass this down in a variable ?

powertools-cloudformation/pom.xml
scottgerring: We should also try run a release locally, without credentials, to make sure that we get to the point where we are pushing to maven (and failing!). I think this should all inherit down, but it would be nice to make sure it does now rather than finding out later :D

examples/pom.xml
jdoherty: You see the following in both cases

Screenshot 2023-12-13 at 14 29 47

Screenshot 2023-12-13 at 14 31 19

Happy to leave it out - but thought it would reduce warnings (but there are lots of other warnings I have not touched that could be fixed such as shading warnings)- up to you!

scottgerring: I agree with you - let's stick with the clean-up!

powertools-e2e-tests/pom.xml
jdoherty: done :-)

scottgerring (approved):
Cheers @jdoherty !

PR #1537, opened 2023-12-12
chore: Support spotbugs running anywhere

scottgerring:
**Issue #, if available:** n/a

Because spotbugs relies on a relative path to find its configuration and license-header, you can't run `mvn install` from arbitrary subdirectories in the project successfully. For instance - `powertools-parameters` - `mvn install` will fail here with an inability to find the `license-header` file. This `maven.multiModuleProjectDirectory` variable only gets set to the actual project root if a [.mvn directory exists](https://stackoverflow.com/questions/29778262/what-is-maven-multimoduleprojectdirectory-used-for). 

By doing this we can easily use it to find files relative to the root without having to specify some relative path property in every single POM. 

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


jeromevdl:
If I understand correctly, it's not only spotbugs but also checkstyle

jeromevdl (approved):

PR #1542, opened 2023-12-22
chore: Remove empty CDK test

scottgerring:
**Issue #, if available:**
The test in the CDK example doesn't really test anything, and adds a requirement on Mac and Windows machines on Docker for regular `mvn install` build. Let's remove it.

The build appears to be broken in `main` so i've fixed that too - see comments, and build history of this branch for details.

## Description of changes:

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


.github/workflows/pr_iac_lint.yml
```diff
@@ -24,6 +24,10 @@ jobs:
         with:
           distribution: 'corretto'
           java-version: 11
+      - name: Build Project
+        working-directory: .
+        run: |
+          mvn install -DskipTests
```
scottgerring, line 30: `terraform validate` needs to be able to find the artifacts of the build process. Because they are all `-SNAPSHOT`, they have to be built. So, build them.

.github/workflows/pr_iac_lint.yml
```diff
@@ -38,7 +42,7 @@ jobs:
           terraform -version
           terraform init -backend=false
           terraform validate
-          terraform plan
+          echo terraform plan # We can't do this without a credentials source
```
scottgerring, line 45: This one needs IAM credentials ?! I've removed it so we do all of the validation apart from the part that connects to AWS. I think this is reasonable.

jeromevdl: OK, let's remove the part that needs AWS credentials, it will avoid failed builds for non-PT-members who push a PR.

jeromevdl, line 45:
```suggestion
```

jeromevdl (approved):

Issue #1523, opened 2023-11-20
RFC: Remove support for Java 8 in V2


scottgerring:
# Removing support of Java 8 in V2

**Date:** November 20, 2023
**Status:** Draft

## Abstract

In this RFC I propose the deprecation of Java 8 support in Powertools for Lambda (Java) starting from version 2 ( #1522 ). Maintaining support for Java 8 has imposed significant resource requirements on the team and limited our ability to focus on new features. While this deprecation is not expected to happen immediately, I am seeking feedback from the community to ensure this is the right path forward with the library. 

## Background

With the release of Java 21 and the ongoing advancements in the Java ecosystem, supporting Java 8 has become increasingly burdensome.

* **Complex Weaving Setup**: Differences in aspect-oriented library support between Java 8 and subsequent versions have become a recurring complication. Powertools must continue to support both the old [AspectJ codehaus plugin from 2021](http://aspectj-maven-plugin/) and current AspectJ and prevents us from using newer features in the library.
* **Friction Adding new Tooling Examples:** Adding Gradle and Kotlin examples needed to take into account the same issues with the AspectJ libraries, increasing frustration of new contributors. For now, we [build Kotlin and Gradle against Java 8 only](https://github.com/aws-powertools/powertools-lambda-java/blob/2602ea7a058d6f644d36bb81e0eabd1ce15ae0fd/.github/workflows/pr_build.yml#L74-L80) because it was too difficult to get them to work generically across this and the other runtimes. 
* **Complex Third Party Library Support:** Mockito has dropped support for Java 8 and we [have to support both old and new versions](https://github.com/aws-powertools/powertools-lambda-java/blob/2602ea7a058d6f644d36bb81e0eabd1ce15ae0fd/pom.xml#L541-L563) increasing the complexity of the build environment and limiting our ability to use newer library features; this will only affect more of our dependencies going forward. Many other popular third party Java libraries have dropped support for Java 8 - Spring Boot 3 (Java 17), Mockito (Java 11). Additionally Kafka 4.0, Flink 2.0, Hadoop, and Spark 4.0 have all dropped Java 8 support. 

This has lead to concrete issues for our users:

[...]

PR #1589, opened 2024-02-29
feat(build): remove java 8 support in v2

jeromevdl:
**Issue #, if available:** #1523

## Description of changes:

Remove java 8 builds for pr and e2e tests github actions in v2.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

not yet a breaking change (just the github actions)

**RFC issue #**:  #1523

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
         with:
           distribution: 'corretto'
-          java-version: 8
+          java-version: 11
           server-id: ossrh
           server-username: MAVEN_USERNAME
           server-password: MAVEN_PASSWORD
--- a/.github/workflows/run-e2e-tests-v2.yml
+++ b/.github/workflows/run-e2e-tests-v2.yml
@@ -1,11 +1,11 @@
-name: Run end-to-end tests
+name: Run end-to-end tests on v2
 
 on:
   workflow_dispatch:
 
   push:
     branches:
-      - main
+      - v2
     paths: # add other modules when there are under e2e tests
       - 'powertools-e2e-tests/**'
       - 'powertools-batch/**'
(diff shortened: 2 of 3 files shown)


scottgerring:
@jeromevdl , we'll need the changes in `.github/workflows/pr_build.yml` in main too. Do you want to raise a PR for that also?

.github/workflows/publish.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
         with:
           distribution: 'corretto'
-          java-version: 8
+          java-version: 11
```
scottgerring, line 16: You're changing this so that when we eventually merge `v2` into main, we change the publish JVM, right? Said otherwise - we don't want to merge this change into `main` too until then (unlike `pr_build.yaml`).

.github/workflows/run-e2e-tests.yml
```diff
@@ -6,7 +6,6 @@ on:
   push:
```
scottgerring, line 6: This one needs to be merged into `main` too

.github/workflows/pr_build.yml
```diff
@@ -1,15 +1,13 @@
-name: Build
```
scottgerring, line 1: This needs to be merged into `main`

.github/workflows/run-e2e-tests-v2.yml
```diff
@@ -0,0 +1,60 @@
...
+      - 'powertools-serialization/**'
+      - 'powertools-tracing/**'
+      - 'pom.xml'
+      - '.github/workflows/**'
+
+  pull_request:
+    branches:
+      - v2
+    paths:
+      - 'powertools-e2e-tests/**'
```
scottgerring, line 28: How come you've removed all the paths here on PR?

.github/workflows/publish.yml
scottgerring: If this is the case, do we also need to change `pr_build` etc as well, so that when we merge v2 into `main` , it ends up setup correctly?

jeromevdl:
replaced by #1606

PR #1606, opened 2024-03-15
feat(build): remove java 8 support in v2

jeromevdl:
**Issue #, if available:** #1523

## Description of changes:

Remove java 8 builds for pr and e2e tests github actions in v2.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

not yet a breaking change (just the github actions)

**RFC issue #**:  #1523

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
         with:
           distribution: 'corretto'
-          java-version: 11
+          java-version: 8
           server-id: ossrh
           server-username: MAVEN_USERNAME
           server-password: MAVEN_PASSWORD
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
         with:
           distribution: 'corretto'
-          java-version: 8
+          java-version: 11
```
scottgerring, line 16: @jeromevdl why are we changing this for v1 / `main` ? seems like a potential problem for little gain - we've never run a publish with a different JDK version, and seeing as we're going to redo the publishing process in v2 anyway probably, I don't see much upside in changing this towards the end of the lifecycle of v1.

scottgerring (changes requested):
A comment on the publish action but otherwise let's get this merged! I'll push it down into v2 straight after.

.github/workflows/publish.yml
jeromevdl: as you prefer. I don't see any risk here but if you want
```suggestion
          java-version: 8
```

Issue #1672, opened 2024-06-26
AppConfig "empty" response handling


chrisclayson:
**What were you trying to accomplish?**

Using the `AppConfigProvider` for the parameters project, I noticed a bug in the handling of instances where AppConfig returns the empty response (i.e. where config has not changed).

This was specifically for a "Freeform configuration profile" rather than feature flags (which I have not tested). 

## Expected Behavior

The provider should re-use the cached config.

## Current Behavior

The provider returns an empty string for the config rather than the cached config. The seems to be because the provider only checks for `null` and not "empty".

## Possible Solution

I implemented a copy of the `AppConfigProvider` class and added a check for `response.configuration().asByteArray().length > 0` and this worked, i.e.:

```java
        // Get the value of the key. Note that AppConfig will return null if the value
        // has not changed since we last asked for it in this session - in this case
        // we return the value we stashed at last request.
        String value = response.configuration() != null && response.configuration().asByteArray().length > 0 ?
                response.configuration().asUtf8String() : // if we have a new value, use it
                establishedSession != null ?
                        establishedSession.lastConfigurationValue :
                        // if we don't but we have a previous value, use that
                        null; // otherwise we've got no value

        // Update the cache so we can get the next value later
        establishedSessions.put(key, new EstablishedSession(nextSessionToken, value));

        return value;
```

I'm happy to submit as a PR.

## Steps to Reproduce (for bugs)

- Setup a freeform configuration profile in AppConfig with some value (we used a JSON object `{ "version": 1 }`)
- Retrieve from a lambda using PowerTools parameters:

```java
          var config = ParamManager.getAppConfigProvider("test", "test")
                .withMaxAge(30, ChronoUnit.SECONDS)
                .get("test");
```
- Fire continuous traffic at the lambda

The first retrieval, on any given runtime environment,  will work fine, but after the 30 second TTL the next retrieve from AppConfg returns an empty string, which is returned from the `get()` method rather than original (cached value).

## Environment

* Powertools for AWS Lambda (Java) 1.18.0
* Built with Gradle
* Java 17 runtime

jeromevdl:
Thank you @chrisclayson for submitting this. And you're more than welcome to create a PR.

@scottgerring, can you have a look?

chrisclayson:
I'll have to fork to submit PR, so I'll give that a go.

[...]

PR #1673, opened 2024-06-26
fix: Allow empty responses as well as null response in AppConfig

chrisclayson:
**Issue #, if available:**

https://github.com/aws-powertools/powertools-lambda-java/issues/1672

## Description of changes:

When checking for empty response from AWS AppConfig data plane, check whether response is an empty byte array rather as well as `null`.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
+++ b/powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
@@ -29,6 +29,8 @@ import software.amazon.lambda.powertools.core.internal.UserAgentConfigurator;
 import software.amazon.lambda.powertools.parameters.cache.CacheManager;
 import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
 
+import static software.amazon.awssdk.utils.StringUtils.isBlank;
+
 /**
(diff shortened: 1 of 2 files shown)


powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
```diff
@@ -29,6 +29,8 @@
 import software.amazon.lambda.powertools.parameters.cache.CacheManager;
 import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
 
+import static software.amazon.awssdk.utils.StringUtils.isBlank;
```
jeromevdl, line 32: Seems it's not used... can we remove.
Plus I'm not sure `blank` is correct, `empty` might be better.

jeromevdl (changes requested):
Almost OK for me, please remove the unused import

powertools-parameters/src/main/java/software/amazon/lambda/powertools/parameters/AppConfigProvider.java
chrisclayson: Oops, left that in from an earlier experiment. Have done a bit of refactoring.

jeromevdl (approved):

Issue #1759, opened 2025-01-31
Bug: v2 StructuredArguments allow logging of reserved keys


phipag:
**What were you trying to accomplish?**
The current `2.0.0-SNAPSHOT` allows logging of reserved keys via structured arguments. This can lead to bugs where users accidentally log a key that is reserved by either the Logging module itself (e.g. `message`) or Powertools.

Examples below are using:
```xml
        <dependency>
            <groupId>software.amazon.lambda</groupId>
            <artifactId>powertools-logging-log4j</artifactId>
            <version>2.0.0-SNAPSHOT</version>
        </dependency>
```

## Expected Behavior
Given the following code, my expected behavior is that the `message` and `service` key appended via `StructuredArguments` are ignored and a user warning is logged stating that the appended keys are reserved keys by Powertools logging.

```java
public class App implements RequestHandler {
    Logger log = LogManager.getLogger(App.class);

    @Logging
    public String handleRequest(final Object input, final Context context) {
        log.info(
                "My message.",
                StructuredArguments.entry("message", "My duplicated message."),
                StructuredArguments.entry("service", "My duplicated service."));
        return "OK";
    }
}
```

Expected output:

```json
{
  "level": "INFO",
  "message": "My message.",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:eu-west-1:<account id>:function:<function name>",
  "function_memory_size": 512,
  "function_name": "<function name>",
  "function_request_id": "9b990378-33dc-470a-a4f2-5f69ccfca154",
  "function_version": "$LATEST",
  "service": "${env:POWERTOOLS_SERVICE_NAME}",
  "timestamp": "2025-01-31T10:54:23.730Z",
  "xray_trace_id": "1-679cabdd-0676a40a4205208f06b16cbf"
}
{
  "level": "WARN",
  "message": "Detected logging of reserved keys 'message', 'service' as structured arguments. Structured arguments with reserved keys will be ignored. More information at <link to docs>.",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:eu-west-1:<account id>:function:<function name>",
  "function_memory_size": 512,
  "function_name": "<function name>",
  "function_request_id": "9b990378-33dc-470a-a4f2-5f69ccfca154",
  "function_version": "$LATEST",
  "service": "${env:POWERTOOLS_SERVICE_NAME}",
  "timestamp": "2025-01-31T10:54:23.730Z",
[...]

PR #1822, opened 2025-04-23
fix(logging): Prevent accidental overwriting of reserved keys via structured arguments

phipag:
**Issue #, if available:** https://github.com/aws-powertools/powertools-lambda-java/issues/1759

## Description of changes:

This PR fixes two problems for both the logback logging module as well as the log4j logging module:
1. Prevents users from accidentally overwriting reserved context keys such as `message`, `level`, Lambda context information, etc.
2. Fixes a bug in the `MapArgument` where JSON serialization was missing a comma if the map contained multiple root-level keys.

Note: In addition to unit tests, I ran E2E tests as well. Until E2E tests are restored in GitHub workflows, here is the output from running in my AWS account.

```
❯ mvn -Pe2e -B verify --file powertools-e2e-tests/pom.xml -Dtest="LoggingE2ET.java"
...
16:32:30.422 [main] INFO  software.amazon.lambda.powertools.testutils.Infrastructure - Uploading asset 3a7b20f10d24fa631c596730d62c85aff6d25420cc2ed78d6a7decf66e08a2e5.jar to cdk-hnb659fds-assets-*******-eu-west-1
16:32:34.738 [main] INFO  software.amazon.lambda.powertools.testutils.Infrastructure - Deploying 'LoggingE2ET-893080a1f7f7' on account *******
16:33:35.840 [main] INFO  software.amazon.lambda.powertools.testutils.Infrastructure - Stack LoggingE2ET-893080a1f7f7 successfully deployed
16:33:38.474 [main] INFO  software.amazon.lambda.powertools.testutils.Infrastructure - Deleting 'LoggingE2ET-893080a1f7f7' on account *******
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 86.50 s -- in software.amazon.lambda.powertools.LoggingE2ET
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

**Checklist**

 

* [x] Update tests
* [x] Update docs

### DIFF
--- a/docs/core/logging.md
+++ b/docs/core/logging.md
@@ -640,6 +640,9 @@ To append additional keys in your logs, you can use the `StructuredArguments` cl
             }
         ```
 
+???+ warning "Warning"
+    If the key name of your structured argument matches any of the [standard structured keys](#standard-structured-keys) or any of the [additional structured keys](#additional-structured-keys) the whole argument will be ignored. This is to protect you from accidentally overwriting reserved keys such as the log level or Lambda context information.
+    
 **Using MDC**
 
 Mapped Diagnostic Context (MDC) is essentially a Key-Value store. It is supported by the [SLF4J API](https://www.slf4j.org/manual.html#mdc){target="_blank"},
--- a/powertools-logging/powertools-logging-log4j/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/PowertoolsResolver.java
+++ b/powertools-logging/powertools-logging-log4j/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/PowertoolsResolver.java
@@ -27,7 +27,9 @@ import static software.amazon.lambda.powertools.logging.internal.PowertoolsLogge
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.logging.log4j.core.LogEvent;
@@ -45,6 +47,9 @@ import software.amazon.lambda.powertools.logging.internal.PowertoolsLoggedFields
  * to be able to recognize powertools fields in the LambdaJsonLayout.json file.
  */
 final class PowertoolsResolver implements EventResolver {
+    private static final Set<String> RESERVED_LOG_KEYS = Stream
+            .concat(PowertoolsLoggedFields.stringValues().stream(), List.of("message", "level", "timestamp").stream())
+            .collect(Collectors.toSet());
 
     private static final EventResolver COLD_START_RESOLVER = new EventResolver() {
         @Override
@@ -191,9 +196,17 @@ final class PowertoolsResolver implements EventResolver {
                     Object[] arguments = logEvent.getMessage().getParameters();
                     if (arguments != null) {
                         stream(arguments).filter(StructuredArgument.class::isInstance).forEach(argument -> {
-                            serializer.writeRaw(',');
                             try {
-                                ((StructuredArgument) argument).writeTo(serializer);
+                                final StructuredArgument structArg = (StructuredArgument) argument;
+                                final Iterable<String> logKeys = structArg.keys();
+                                // If any of the logKeys are a reserved key we are going to ignore the argument
+                                for (final String logKey : logKeys) {
+                                    if (RESERVED_LOG_KEYS.contains(logKey)) {
+                                        return;
+                                    }
+                                }
+                                serializer.writeRaw(',');
+                                structArg.writeTo(serializer);
                             } catch (IOException e) {
                                 System.err.printf("Failed to encode log event, error: %s.%n", e.getMessage());
                             }
--- a/powertools-logging/powertools-logging-log4j/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/PowertoolsResolver.java
+++ b/powertools-logging/powertools-logging-log4j/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/PowertoolsResolver.java
@@ -48,7 +48,8 @@ import software.amazon.lambda.powertools.logging.internal.PowertoolsLoggedFields
  */
 final class PowertoolsResolver implements EventResolver {
     private static final Set<String> RESERVED_LOG_KEYS = Stream
-            .concat(PowertoolsLoggedFields.stringValues().stream(), List.of("message", "level", "timestamp").stream())
+            .concat(PowertoolsLoggedFields.stringValues().stream(),
+                    List.of("message", "level", "timestamp", "error").stream())
             .collect(Collectors.toSet());
 
     private static final EventResolver COLD_START_RESOLVER = new EventResolver() {
--- a/powertools-logging/powertools-logging-log4j/src/test/java/org/apache/logging/log4j/layout/template/json/resolver/PowertoolsResolverArgumentsTest.java
+++ b/powertools-logging/powertools-logging-log4j/src/test/java/org/apache/logging/log4j/layout/template/json/resolver/PowertoolsResolverArgumentsTest.java
@@ -35,6 +35,8 @@ import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mock;
 import org.slf4j.MDC;
+
+import software.amazon.lambda.powertools.logging.internal.PowertoolsLoggedFields;
 import software.amazon.lambda.powertools.logging.internal.handler.PowertoolsArguments;
 
 @Order(2)
@@ -83,9 +85,14 @@ class PowertoolsResolverArgumentsTest {
         // THEN
         File logFile = new File("target/logfile.json");
         assertThat(contentOf(logFile))
-                .contains("\"input\":{\"awsRegion\":\"eu-west-1\",\"body\":\"plop\",\"eventSource\":\"eb\",\"messageAttributes\":{\"keyAttribute\":{\"stringListValues\":[\"val1\",\"val2\",\"val3\"]}},\"messageId\":\"1212abcd\"}")
+                .contains(
+                        "\"input\":{\"awsRegion\":\"eu-west-1\",\"body\":\"plop\",\"eventSource\":\"eb\",\"messageAttributes\":{\"keyAttribute\":{\"stringListValues\":[\"val1\",\"val2\",\"val3\"]}},\"messageId\":\"1212abcd\"}")
                 .contains("\"message\":\"1212abcd\"")
                 .contains("\"message\":\"Message body = plop and id = \\\"1212abcd\\\"\"");
+        // Reserved keys should be ignored
+        PowertoolsLoggedFields.stringValues().stream().forEach(reservedKey -> {
+            assertThat(contentOf(logFile)).doesNotContain("\"" + reservedKey + "\":\"shouldBeIgnored\"");
+        });
     }
 
     @Test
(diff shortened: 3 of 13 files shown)


docs/core/logging.md
```diff
@@ -640,6 +640,9 @@ To append additional keys in your logs, you can use the `StructuredArguments` cl
             }
         ```
 
+???+ warning "Do not use arguments with reserved keys"
```
dreamorosi, line 643: I would consider just adding a notice/warning directly in [this section](https://docs.powertools.aws.dev/lambda/java/core/logging/#standard-structured-keys) so that customers don't need to jump around, similar to what we did in TypeScript [see wording](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/#standard-structured-keys).

However I also noticed that the table for structured keys in the docs might need to be updated. In the code you correctly set as "reserved" keys some of them, but the table mixes other keys from the context that are not standard keys (i.e. `functionName`, `functionVersion`, etc.). Perhaps better split that table and add these under the section where you talk about context instead.

docs/core/logging.md
```diff
@@ -650,6 +653,9 @@ Mapped Diagnostic Context (MDC) is essentially a Key-Value store. It is supporte
 ???+ warning "Custom keys stored in the MDC are persisted across warm invocations"
     Always set additional keys as part of your handler method to ensure they have the latest value, or explicitly clear them with [`clearState=true`](#clearing-state).
 
+???+ warning "Do not add reserved keys to MDC"
```
dreamorosi, line 656: I googled MDC, but could you share more as for why we are not applying this same key filtering logic to keys set to MDC?

Also the product name is `Powertools for AWS`; if you don't want to use this just refer to it as `Logger` in this type of context.

dreamorosi (changes requested):
Left a couple of comments on the documentation, one of which needs attention.

Regarding the implementation, two comments:
- in TS and I think other runtimes, we are emitting a warning when we ignore a key to let the customer know there's been data loss - see [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/114188cd2bf230c59570f5fa05c552cbe2ed6d3d/packages/logger/src/Logger.ts#L968-L974). Was this omission intentional?
- I don't know the implementation good enough to answer this question without running the Logger, so I'll ask: if I add a reserved key via the [`appendKey()` method](https://docs.powertools.aws.dev/lambda/java/core/logging/#appending-additional-keys), is it being correctly ignored? 

Finally, could you look at the SonarQubeCloud findings and either address them or mark them as false positives?

hjgraca:
> > in TS and I think other runtimes, we are emitting a warning when we ignore a key to let the customer know there's been data loss - see [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/114188cd2bf230c59570f5fa05c552cbe2ed6d3d/packages/logger/src/Logger.ts#L968-L974). Was this omission intentional?
> 
> Yes, it was intentional because I didn't want to spam the user's logs with warning messages on each log emission where this happens. If you have a strong opinion on adding a warning I can do so. I believe in .NET there is a warning in the documentation only – similar to what I propose here.

Correct, we do not log any warning to user.

docs/core/logging.md
phipag: I will work on this change and send a new commit.

Regarding the table, I believe this is already fixed in the preview documentation, did you look at the `v1` docs? Here is a link to the preview docs tables. There it is split into two tables: https://docs.powertools.aws.dev/lambda/java/preview/core/logging/#standard-structured-keys

Is this what you mean?

dreamorosi:
Thanks for the explanations.

As a customer, if I am emitting a log with a key and that key is dropped, I would want to know. Putting myself in the shoes of an operator, having a warning that tells me this has happened helps me understand.

The opposite means: 1/ I see a missing key from my logs when I needed it, 2/ I think there's an issue with my code, 3/ I spend time troubleshooting, 4/ maybe, after a long while, I understand this is because of a Powertools specific behavior, 5/ I'm now frustrated because I wasted time.

Since we are saying we're excluding MDC from this logic (more on this in a sec), then we'd not be spamming the customer since the warning would be local to a specific log entry (aka it's emitted once). Besides, warnings are subject to log level, so I could silence them if I really don't want them.

Alternatively, we could also have a "warn once" logic that emits a warning at most once whenever a key filtering action happens the first time. This way we still inform the customer, but keep warnings at minimum. This is still a better option than saying nothing in my opinion.

Regarding MDC, this is again a limitation of knowledge in the language/ecosystem, but does this code (from the docs) add the keys to that one specific log entry, or does it add to all the subsequent ones?

```java
import static software.amazon.lambda.powertools.logging.argument.StructuredArguments.entry;
import static software.amazon.lambda.powertools.logging.argument.StructuredArguments.entries;

public class PaymentFunction implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {

    private static final Logger LOGGER = LoggerFactory.getLogger(AppLogResponse.class);

    @Logging
    public APIGatewayProxyResponseEvent handleRequest(final APIGatewayProxyRequestEvent input, final Context context) {
        // ...
        LOGGER.info("Collecting payment", entry("orderId", order.getId()));

        // ...
        Map<String, String> customKeys = new HashMap<>();
        customKeys.put("paymentId", payment.getId());
        customKeys.put("amount", payment.getAmount);
        LOGGER.info("Payment successful", entries(customKeys));
    }
}
```

phipag:
> Regarding MDC, this is again a limitation of knowledge in the language/ecosystem, but does this code (from the docs) add the keys to that one specific log entry, or does it add to all the subsequent ones?

This code will add the keys only to those specific log entries. If you want to add keys to all log entires you can do this – and this where we can't prevent users from adding reserved keys:

```java
public class CreditCardFunction implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {

    private static final Logger LOGGER = LoggerFactory.getLogger(CreditCardFunction.class);

    @Logging(clearState = true)
    public APIGatewayProxyResponseEvent handleRequest(final APIGatewayProxyRequestEvent input, final Context context) {
        // This will be attached to each log
        MDC.put("cardNumber", card.getId());
        LOGGER.info("Updating card information");
        // ...
    }
}
```

> Besides, warnings are subject to log level, so I could silence them if I really don't want them.

This is a strong argument, I can send an update emitting a warning log.

docs/core/logging.md
phipag: I sent an update addressing this comment.

docs/core/logging.md
phipag: I updated the docs and deployed a preview here: https://dealn7fl31ram.cloudfront.net/core/logging/#standard-structured-keys.

I added the same note like in TypeScript right at the beginning where the documentation introduces structured keys.

I kept the warnings in the respective section of `StructuredArguments` and `MDC` with the details on how each method is treated.

phipag:
Hey @hjgraca and @dreamorosi,

thank you again for your comments. I sent an update addressing your comments. While working on this I also improved the implementation by moving the reserved key logic out of the respective log4j/logback implementation to the `StructuredArguments` class which removes a lot of redundant code. This also allowed making each implementation of a structured argument (e.g. `MapArgument`) package private reducing the public API surface area.

Let me summarize the current behavior with an example:

#### `StructuredArguments`

```java
    @Logging(clearState = true)
    public String handleRequest(final Object input, final Context context) {
        LOGGER.info("My message.", StructuredArguments.entry("message", "My duplicated message."),
                StructuredArguments.entry("message2", "My second message."),
                StructuredArguments.entry("service", "My duplicated service."),
                StructuredArguments.entry("function_name", "My duplicated function_name"));
        return "OK";
    }
```

As discussed, this will drop the reserved keys (`message`, `service`, `function_name`) and log exactly one warning for each offending key. This warning can be suppressed by the user when filtering by log-level.

```json
{
  "level": "WARN",
  "message": "Attempted to use reserved key 'message' in structured argument. This key will be ignored.",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:us-east-1:012345678912:function:HelloWorldFunction",
  "function_memory_size": 512,
  "function_name": "HelloWorldFunction",
  "function_request_id": "03f6ec1f-51dd-4a63-87de-8831ca1fd883",
  "function_version": "$LATEST",
  "service": "service_undefined",
  "timestamp": "2025-04-24T10:54:40.120Z"
}
{
  "level": "WARN",
  "message": "Attempted to use reserved key 'service' in structured argument. This key will be ignored.",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:us-east-1:012345678912:function:HelloWorldFunction",
  "function_memory_size": 512,
  "function_name": "HelloWorldFunction",
  "function_request_id": "03f6ec1f-51dd-4a63-87de-8831ca1fd883",
  "function_version": "$LATEST",
  "service": "service_undefined",
  "timestamp": "2025-04-24T10:54:40.121Z"
}
{
  "level": "WARN",
  "message": "Attempted to use reserved key 'function_name' in structured argument. This key will be ignored.",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:us-east-1:012345678912:function:HelloWorldFunction",
  "function_memory_size": 512,
  "function_name": "HelloWorldFunction",
  "function_request_id": "03f6ec1f-51dd-4a63-87de-8831ca1fd883",
  "function_version": "$LATEST",
  "service": "service_undefined",
  "timestamp": "2025-04-24T10:54:40.122Z"
}
{
  "level": "INFO",
  "message": "My message.",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:us-east-1:012345678912:function:HelloWorldFunction",
  "function_memory_size": 512,
  "function_name": "HelloWorldFunction",
  "function_request_id": "03f6ec1f-51dd-4a63-87de-8831ca1fd883",
  "function_version": "$LATEST",
  "service": "service_undefined",
  "timestamp": "2025-04-24T10:54:40.123Z",
  "message2": "My second message."
}
```

#### Mapped Diagnostic Context `MDC`

Here, we have the technical limitation of not being able to prevent the user from doing this and therefore added a warning about this behavior in the documentation.

```java
    @Logging(clearState = true)
    public String handleRequest(final Object input, final Context context) {
        MDC.put("message", "My duplicated message.");
        LOGGER.info("My message.");
        return "OK";
    }
```

This code will render the `message` key twice in the JSON output:

```json
{"level":"INFO","message":"My message.","cold_start":true,"function_arn":"arn:aws:lambda:us-east-1:012345678912:function:HelloWorldFunction","function_memory_size":512,"function_name":"HelloWorldFunction","function_request_id":"e7a2a7f2-b9db-49d4-929f-72a89f002334","function_version":"$LATEST","service":"service_undefined","timestamp":"2025-04-24T11:06:32.413Z","message":"My duplicated message."}
```

which will yield the following parsed JSON (in CloudWatch):

```json
{
  "level": "INFO",
  "message": "My duplicated message.",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:us-east-1:012345678912:function:HelloWorldFunction",
  "function_memory_size": 512,
  "function_name": "HelloWorldFunction",
  "function_request_id": "e7a2a7f2-b9db-49d4-929f-72a89f002334",
  "function_version": "$LATEST",
  "service": "service_undefined",
  "timestamp": "2025-04-24T11:06:32.413Z"
}
```

dreamorosi (approved):
Thanks for addressing my feedback and adding the warnings.

This MDC thing is unfortunate, and I wish there was a better solution, but for now we'll keep it as it is.

Issue #1588, opened 2024-02-29
Add support for CRaC


scottgerring:
**Is your feature request related to a problem? Please describe.**
We should add support for CRaC priming. This will improve the startup performance of applications using Powertools. @msailes has written a helpful guide [here](https://github.com/marksailes/snapstart-priming-guide/). I think we should do this before v2 is released.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

subhash686:
@scottgerring I would like to take a CRaCk at this :) 
Given this is an enhancement, should I create any additional documentation before starting? Also what is a potential v2 release date that I can work backwards from?


Describe the solution you'd like
Automatic priming of sub-modules independently helps the run time snapshot only what is required. 
Try to create a list of classes at build time before packaging (Need to research this a little, but we can definitely start with a manual step of generating such a list)
Preload the identified classes using CRaC 

Describe alternatives you've considered
Manual priming may be simpler but we need to decide good code-paths to hint JIT. These are subjective and need maintenance when things change.

Additional context
@msailes  blog and examples from embedded metrics

scottgerring:
Hey @subhash686 glad you're interested! I'm no longer a maintainer here, but:

@msailes @dreamorosi @jeromevdl can anyone point @subhash686 in the right direction?

Vadym79:
I think SnapStart priming can be done in Powertools for several use cases:
1) Powertools primes itself automatically. Everything which initialized behind Powertools annotations like @Idempotency (for example DynamoDB PersistentStore), @Logging @Metrics and @Tracing can be primed, otherwise the performance with Powertools is not so great. It too much effort to do things like [this](https://github.com/Vadym79/AWSPowertoolsForLambdaJavaV2/blob/main/java21-lambda-idempotency/src/main/java/software/amazonaws/example/product/handler/CreateProductWithDynamoDBIdempotencyPrimingHandler.java) and [this](https://github.com/Vadym79/AWSPowertoolsForLambdaJavaV2/blob/main/java21-lambda-observability/src/main/java/software/amazonaws/example/product/handler/GetProductByIdWithObservabilityPrimingHandler.java) by myself.

[...]

PR #1861, opened 2025-05-28
feat: Support CRaC and priming of powertools metrics and idempotency-dynamodb

subhash686:
**Issue #, if available:**
https://github.com/aws-powertools/powertools-lambda-java/issues/1588

## Description of changes:

- Added CRaC support in powertools-common, following the guide and blogs mentioned on the ticket.
- Added automatic priming for powertools-metrics by preloading all the required classes, the list of classes is fetched from the JVM at compile time.
- Added manual priming for dynamodb persistence store

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


subhash686:
@phipag Looks like v2 is now merged and released, so I resolved conflicts to rebase to `main`. Let me know if you have any other release plans for this.

phipag:
Thanks @subhash686 and apologize the delay. I was very busy with the v2 release. 

Your PR is next on my list and we'll try to get it merged for the next minor release 2.1.0.

phipag:
Hey @subhash686, I didn't forget you. This week is very busy but I'll do my best to test your implementation very soon. 

In the meantime, can you have a look at the 4 Sonar issued and either resolve them or mark them as not applicable if you think that is the case?

subhash686:
@phipag Fixed the sonar issues, I did not notice the comment earlier, thanks for pointing it out.

phipag:
Hey @subhash686 , I am reviewing your PR now. Can you provide some details how you generated the txt file of the classes loaded? I see you say:

> the list of classes is fetched from the JVM at compile time.

Can you provide the compilation command you used for this?

Ideally, I would like to reproduce the process for how to generate and then use the txt file of loaded classes.

subhash686:
> Hey @subhash686 , I am reviewing your PR now. Can you provide some details how you generated the txt file of the classes loaded? I see you say:
> 
> > the list of classes is fetched from the JVM at compile time.
> 
> Can you provide the compilation command you used for this?
> 
> Ideally, I would like to reproduce the process for how to generate and then use the txt file of loaded classes.

Thanks @phipag . Here are the steps I followed
- Ran following commands to generate the loaded classes in the project root directory
`> export MAVEN_OPTS="-Xlog:class+load=info:classloaded.txt"`
`> mvn install -pl powertools-metrics -am` 

- File has each line in the format `[0.054s][info][class,load] java.lang.Object source: shared objects file`
- Used regex to remove all the prefix extra characters `^\[[\[\]0-9.a-z,]+ ` and replace the source text using regex`( source: )[0-9a-z:/._$-]+` which is at the end of each line.
- Moved the file to powertools-metrics `resources` directory.

phipag:
Thanks @subhash686 for providing this context. I will reproduce it and run a load test of this code. Will share the results here soon and we can move forward.

phipag:
Hey @subhash686,

here is my first update. I tested the classesload.txt (automatic priming) implementation that you contributed and found some issues:
1. This is not running when included in the `LambdaMetricsAspect` that is annotated with `@Aspect`. I moved the priming code for loading the classes into the [EmfMetricsLogger.java](https://github.com/aws-powertools/powertools-lambda-java/blob/8a040ac87a2a31662b7ac8980601b4d821cc29d8/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java) which worked! I verified this by adding a log statement in the beforeCheckpoint and afterRestore methods and verified in a Snapstart deployment that this gets logged.
2. The classesloaded.txt file does not represent the classes loaded at runtime. Since you generated the file during `mvn install -pl powertools-metrics -am` this will yield the loaded classes during all maven stages part of the install goal. This does not achieve the intended goal which is to pre-load all classes that are essential at runtime of the application. For example, if you search in the file for "MetricsFactory" nothing will show up but this class is mandatory to be loaded at runtime when using the Metrics module.
    - What we do for GraalVM is documented [here](https://github.com/aws-powertools/powertools-lambda-java/blob/main/GraalVM.md). We run the GraalVM tracing agent as part of the unit tests. This way we can actually get the metadata for the loaded classes at runtime (or at least an approximation of it). I envision we can do the same thing here by tracing the loaded classes during execution of the unit tests.  
    - If you checkout @msailes example [here](https://github.com/marksailes/snapstart-priming-guide/blob/main/examples/emf/emf-client-example/src/main/resources/classes.txt) it will have those classes like `MetricsLogger` etc.

For issue (2), I think this needs a completely different approach for generating the classesloaded.txt. We need to implement a mechanism to capture the classes at runtime not the classes used by maven during compilation. Possibly, the GraalVM mechanism can be re-used in some form?

I will now review the Idempotency manual priming that you implemented and will get back to you again. Just wanted to share these preliminary findings. Let me know if you have any suggestion / idea about this. I am happy to join a meeting with you to brainstorm more?

subhash686:
Hi @phipag 
Thanks so much for the time you spent adjusting the code and thorough suggestions. 

For (2), I created a lambda handler with FlushMetrics annotation and used it in a test to generated the classes again. Used GraalVM profile to run all tests with the -Xlog JVM option. I see the runtime classes now with some extra test classes.
I replaced the classesloaded file for now, so we can continue testing. I am happy to meet to brainstorm the approach to generate the class list. 

I will fix (1) once you get a chance to review it all.

powertools-idempotency/powertools-idempotency-core/src/main/java/software/amazon/lambda/powertools/idempotency/internal/IdempotencyHandler.java
```diff
@@ -63,6 +64,28 @@ public IdempotencyHandler(ProceedingJoinPoint pjp, String functionName, JsonNode
...
+        } catch (Exception unknown) {
+            // This is unexpected but we must continue without any interruption
+        }
+    }
+
+    @Override
+    public void afterRestore(org.crac.Context<? extends Resource> context) throws Exception {
+        // This is a no-op, as we don't need to do anything after restore
+    }
+
```
phipag: Can we move this to the DynamoDBPersistenceStore.java? It will not run when included here because this class is only used from the AspectJ Aspect. 

Also, don't forget to register the CraC resource in the constructor of the class.

From the DynamoDBPersistenceStore we can then run some more logic such as createRecord, updateRecord, and getRecord:

```java
    /**
     * Primes the persistent store by invoking the get record method with a key that doesn't exist.
     *
     * @param context
     * @throws Exception
     */
    @Override
    public void beforeCheckpoint(org.crac.Context<? extends Resource> context) throws Exception {
        try {
            System.out.println("beforeCheckpoint");
            this.getRecord("__invoke_prime__");
            // Add logic for createRecord, updateRecord here and catch exceptions
        } catch (IdempotencyItemNotFoundException keyNotFound) {
            // This is expected, as we are priming the store
            System.out.println("beforeCheckpoint IdempotencyItemNotFoundException");
        } catch (Exception unknown) {
            // This is unexpected but we must continue without any interruption
            System.out.println("beforeCheckpoint Exception");
        }
    }

    @Override
    public void afterRestore(org.crac.Context<? extends Resource> context) throws Exception {
        // This is a no-op, as we don't need to do anything after restore
        System.out.println("afterRestore");
    }
```

powertools-idempotency/pom.xml
```diff
@@ -46,6 +46,10 @@
             <groupId>software.amazon.lambda</groupId>
             <artifactId>powertools-common</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.crac</groupId>
+            <artifactId>crac</artifactId>
+        </dependency>
```
phipag: Let's also move this dependency to the dynamodb module for now.

phipag:
Hey @subhash686,

I reviewed and tested the DynamoDB priming (see my comments above). I made these changes locally and ran a load test with your priming code and without the priming code. I observe positive results that confirm priming actually improves the restore duration. Here are my results:

### Idempotency Snapstart restore duration 
Tested using example at https://github.com/aws-powertools/powertools-lambda-java/blob/984036ca2e3afb92b2c838bcfa2e48a51c3d7a64/examples/powertools-examples-idempotency

Priming achieves ~27% better restore duration.
#### Without priming

| snapstart_restore_count | avg_restore_ms | min_restore_ms | max_restore_ms | p50_restore_ms | p90_restore_ms | p95_restore_ms | p99_restore_ms |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 401 | 673.2549 | 389.51 | 1081.39 | 675.0206 | 766.3812 | 820.28 | 894.8018 |
#### With priming

| snapstart_restore_count | avg_restore_ms | min_restore_ms | max_restore_ms | p50_restore_ms | p90_restore_ms | p95_restore_ms | p99_restore_ms |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 402 | 440.2612 | 177.07 | 719.34 | 438.3243 | 528.9348 | 556.596 | 655.0805 |

---

>For (2), I created a lambda handler with FlushMetrics annotation and used it in a test to generated the classes again. Used GraalVM profile to run all tests with the -Xlog JVM option. I see the runtime classes now with some extra test classes.
I replaced the classesloaded file for now, so we can continue testing.

Nice, the file looks much better now although it also includes the unit test libraries like `org.junit` which is expected. We have the same problem in GraalVM and rely on manually cleaning this up. In this case, it might not be a big deal because during the `beforeCheckpoint` hook, we will simply ignore classes that are not found at runtime. 

> I am happy to meet to brainstorm the approach to generate the class list.

Happy to hear that. Let's setup a meeting to brainstorm the best approach together for this class loading. Can you send an email to `aws-powertools-maintainers@amazon.com` and share your timezone and availability (I am based in CET timezone)? I will send out a meeting invite then 🚀

phipag:
I performed the same load test for the Classloading using the new txt file you provided @subhash686. Here are my results:

### Metrics priming (Classloading)

Tested using example at https://github.com/aws-powertools/powertools-lambda-java/blob/984036ca2e3afb92b2c838bcfa2e48a51c3d7a64/examples/powertools-examples-core-utilities/sam (removed tracer and logger).

We can see ~7% better performance using classloading priming.

#### Without priming

| snapstart_restore_count | avg_restore_ms | min_restore_ms | max_restore_ms | p50_restore_ms | p90_restore_ms | p95_restore_ms | p99_restore_ms |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 400 | 634.9481 | 443.29 | 966.02 | 623.1474 | 730.4813 | 778.7357 | 894.8018 |

#### With priming

| snapstart_restore_count | avg_restore_ms | min_restore_ms | max_restore_ms | p50_restore_ms | p90_restore_ms | p95_restore_ms | p99_restore_ms |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 405 | 586.1477 | 345.14 | 944.56 | 581.0391 | 678.4024 | 718.1742 | 831.0078 |

phipag:
Next steps:
- Address issues (1) and (2)
- Brainstorm in a meeting (@phipag and @subhash686) to understand best process for generating classloaded.txt files
- Address PR comments (move Idempotency priming logic to DDB datastore and increase priming scope)
- Merge PR

phipag:
Thanks for joining the meeting @subhash686! Here is a short summary of what we discussed:

- Use a maven profile to auto-generate the classesloaded.txt files. Something similar to this:

```xml
        <profile>
            <id>generate-classesloaded-file</id>
            <build>
                <plugins>
                    <plugin>
                        <groupId>org.apache.maven.plugins</groupId>
                        <artifactId>maven-surefire-plugin</artifactId>
                        <version>3.2.3</version>
                        <configuration>
                            <argLine>-Xlog:class+load=info:classloaded.txt</argLine>
                        </configuration>
                    </plugin>
                </plugins>
            </build>
        </profile>
```

It is okay if unit test classes are in there, we ignore them in the `ClassPreLoader` and the `beforeCheckpoint` hook is not time sensitive.

- Move the priming logic in Metrics to the `MetricsFactory` (AspectJ is not considered for priming)
- Move the priming logic for Idempotency to the DynamoDb persistence store (This will allow us to prime data store specifics)
- Increase the priming scope in the DynamoDB persistence store by introducing putRecord, getRecord, deleteRecord calls. This should prime the success scenario of the persistence store potentially covering more code branches.
- Create a Priming.md documentation explaining the process and how to run it and generate the classesloaded.txt file similar to [GraalVM.md](https://github.com/aws-powertools/powertools-lambda-java/blob/main/GraalVM.md)
- Optional: Research a potential automation of generating the classesload.txt file via a GitHub actions that creates a Pull Request if the file changed after introducing code changes. This will make sure, that in the future we keep the files up-to-date.

subhash686:
@phipag It was great meeting with you. 
- I moved the class priming to `MetricsFactory`
- Moved invoke priming to `DynamoDBPersistenceStore` and extended the scope to add and delete record as well.
- Added a new maven profile to generate classes
- Documented it all in  Priming.md

phipag:
Awesome, thanks @subhash686. I'll review this by the end of the week.

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+- [Overview](#overview)
+- [Implementation Steps](#general-implementation-steps)
+- [Known Issues](#known-issues-and-solutions)
+- [Reference Implementation](#reference-implementation)
+
+## Overview
+Priming is the process of preloading dependencies and initializing resources during the INIT phase, rather than during the INVOKE phase to further optimize startup performance with SnapStart. 
+This is required because Java frameworks that use dependency injection load classes into memory when these classes are explicitly invoked, which typically happens during Lambda’s INVOKE phase.
+
+This documentation provides guidance for automatic class priming in AWS Lambda Powertools Java modules.
```
phipag:
```suggestion
This documentation provides guidance for automatic class priming in Powertools for AWS Lambda Java modules.
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+## Implementation Steps
+Classes are proactively loaded using Java runtime hooks which are part of the open source CRaC (Coordinated Restore at Checkpoint) project.
```
phipag:
```suggestion
Classes are proactively loaded using Java runtime hooks which are part of the open source [CRaC (Coordinated Restore at Checkpoint) project](https://openjdk.org/projects/crac/).
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+This implementation uses this hook, called `beforeCheckpoint()`, to prime SnapStart-enabled Java functions via Class Priming.
```
phipag:
```suggestion
Implementations across the project use the `beforeCheckpoint()` hook, to prime Snapstart-enabled Java functions via Class Priming.
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+In order to generate the `classloaded` files for Powertools java module, follow these general steps.
```
phipag:
```suggestion
In order to generate the `classesloaded.txt` file for a Java module in this project, follow these general steps.
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. **Add Maven Profile**
+    - Add maven test profile with the following VM argument for generating classes loaded files. 
+   ```shell
+      -Xlog:class+load=info:classesloaded.txt
+      ```
+    - You can find an example of this in profile `generate-classesloaded-file` in this [pom.xml](powertools-metrics/pom.xml).
```
phipag:
```suggestion
    - You can find an example of this `generate-classesloaded-file` profile in this [pom.xml](powertools-metrics/pom.xml).
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+    - Run tests with `-Pgenerate-classesloaded-file` profile.
+   ```shell
+   mvn -Pgenerate-classesloaded-file clean test
+   ```
+   - This will generate a file named `classesloaded.txt` in the target directory of the module.
+
+3. **Cleanup the file**
+    - The classes loaded file generated in Step 2 has the format
+   `[0.054s][info][class,load] java.lang.Object source: shared objects file` 
+   but we are only interested in `java.lang.Object` the fully qualified class name.
```
phipag:
```suggestion
   but we are only interested in `java.lang.Object` – the fully qualified class name.
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+    - To strip the lines to include only the fully qualified class name, 
+      Use the following regex to replace with empty string.
+      - `^\[[\[\]0-9.a-z,]+ ` (to replace the left part)
+      - `( source: )[0-9a-z :/._$-]+` (to replace the right part)
+
+4. **Place the file**
```
phipag:
```suggestion
4. **Add file to resources**
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+    -  Move the cleanedup file containing to the corresponding resources directory of the module. See [example](powertools-metrics/src/resources/classesloaded.txt).
```
phipag:
```suggestion
    -  Move the cleaned-up file to the corresponding `src/main/resources` directory of the module. See [example](powertools-metrics/src/main/resources/classesloaded.txt).
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+
+5. **Register and checkpoint**
```
phipag, line 47: Let's clarify this section by stating the following information:
- Mention that ClassPreLoader is implemented in `powertools-common` module
- Mention that this cannot be done in an AspectJ aspect
- Mention that this needs to be done in a suitable class. A class is suitable if it is the "entrypoint" for a module – i.e. `MetricsFactory` for Metrics module.

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+    - Register the CRaC Resource in the constructor of the main class.
+    - Add the `beforeCheckpoint()` hook in the same class to invoke `ClassPreLoader.preloadClasses()`.
+    - This will ensure that the classes are loaded during the INIT phase of the Lambda function. See [example](powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java)
```
phipag:
```suggestion
    - This will ensure that the classes are already pre-loaded by the Snapstart RESTORE operation leading to a shorter INIT duration. See [example](powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java)
```

Priming.md
```diff
@@ -0,0 +1,58 @@
...
+  
+
+## Known Issues
+- This is a manual process at the moment, but it can be automated in the future.
+- Because the file is generated while running tests, it includes test classes as well.
```
phipag: Let's explain here why this is not a problem. It is not a problem because all classes that are not found are ignored by the ClassPreLoader and the `beforeCheckpoint()` hook is not time sensitive since it only runs once when a new Lambda function version gets published.

phipag (changes requested):
Thanks @subhash686. I added some cosmetic comments in the Priming documentation. I am deploying your changes to a Snapstart Lambda for testing now. I will get back with the results later.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
```diff
@@ -23,11 +26,12 @@
 /**
  * Factory for accessing the singleton Metrics instance
  */
-public final class MetricsFactory {
+public final class MetricsFactory implements Resource {
     private static MetricsProvider provider = new EmfMetricsProvider();
     private static Metrics metrics;
 
     private MetricsFactory() {
+        Core.getGlobalContext().register(this);
```
phipag: This doesn't work since this class has a private constructor. It is a static utility class.

Instead, we should use a static block to register a static dummy instance of that class to enable CRaC for this class. Like this:


```java
    // Create field holding dummy instance
    private static final MetricsFactory INSTANCE = new MetricsFactory();

   // Static block registering MetricsFactory with CRaC 
   static {
        Core.getGlobalContext().register(INSTANCE);
    }
```

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
```diff
@@ -68,4 +72,14 @@ public static synchronized void setMetricsProvider(MetricsProvider metricsProvid
         // Reset the metrics instance so it will be recreated with the new provider
         metrics = null;
     }
+
+    @Override
+    public void beforeCheckpoint(org.crac.Context<? extends Resource> context) throws Exception {
+        ClassPreLoader.preloadClasses();
```
phipag, line 83: We can take the opportunity here to call `MetricsFactory.getMetricsInstance()` to prime getting an instance. This will always be called statically and count into the cold start otherwise.

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.net.URL;
+import java.net.URLConnection;
+import java.util.Enumeration;
+
+/**
+ * Used to preload classes to support automatic priming for SnapStart
+ */
+public class ClassPreLoader {
```
phipag: You can make this class final since it only has private constructors.

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public static void preloadClasses() {
+        try {
+            Enumeration<URL> files = ClassPreLoader.class.getClassLoader().getResources(CLASSES_FILE);
+            // If there are multiple files, preload classes from all of them
+            while (files.hasMoreElements()) {
+                URL url = files.nextElement();
+                URLConnection conn = url.openConnection();
+                conn.setUseCaches(false);
+                InputStream is = conn.getInputStream();
```
phipag, line 46: Make sure the input stream gets closed properly using a `try(InputStream is = ...) {...}` block.

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
```diff
@@ -0,0 +1,77 @@
...
+            String line;
+            while ((line = reader.readLine()) != null) {
+                int idx = line.indexOf('#');
+                if (idx != -1) {
+                    line = line.substring(0, idx);
+                }
+                final String className = line.stripTrailing();
+                if (!className.isBlank()) {
+                    Class.forName(className, true, ClassPreLoader.class.getClassLoader());
+                }
```
phipag: I found a bug in this logic. We exit this while loop when the first class cannot be found. We should catch the exception raised by `Class.forName` and `continue` instead to allow loading of all classes. 

I tested it using this code (you can remove the `System.out.println`):
```java
                    try {
                        System.out.println("ClassPreLoader loading " + className);
                        Class.forName(className, true, ClassPreLoader.class.getClassLoader());
                    } catch (Exception e) {
                        System.out.println("ClassPreLoader ignored " + e.getMessage());
                        continue;
                    }
```

Here is an example of the logs I checked. You can see it loads MetricsFactory correctly but ignores the Test classes.

Image

powertools-idempotency/powertools-idempotency-dynamodb/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/dynamodb/DynamoDBPersistenceStore.java
```diff
@@ -52,7 +55,7 @@
  * DynamoDB version of the {@link PersistenceStore}. Will store idempotency data in DynamoDB.<br>
  * Use the {@link Builder} to create a new instance.
  */
-public class DynamoDBPersistenceStore extends BasePersistenceStore implements PersistenceStore {
+public class DynamoDBPersistenceStore extends BasePersistenceStore implements PersistenceStore, Resource {
```
phipag: This class can be final as well since it only has a private constructor and uses the Builder pattern.

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.junit.jupiter.api.Test;
+
+class ClassPreLoaderTest {
+
+    @Test
+    void preloadClasses_shouldIgnoreInvalidClassesAndLoadValidClasses() {
+        // Verify that the missing class did not throw any exception
+        assertDoesNotThrow(ClassPreLoader::preloadClasses);
+    }
```
phipag: **Optional:** We should add more thorough unit tests for the class pre loading with a mock classesloaded.txt file in the src/test/resources folder. 

Do not prioritize this, let's focus on implementing the other comments first.

Priming.md
phipag: See this comment. There is a bug currently that we need to fix to ignore classes that failed to load https://github.com/aws-powertools/powertools-lambda-java/pull/1861#discussion_r2240091334

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
subhash686: I have the try-with-resources statement in the method below which ensures proper closing. Do you see any concern with that approach?

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Added language to clarify entrypoint selection to register with CRaC.

Priming.md
subhash686: Modified per the suggestion.

Priming.md
subhash686: Modified per the suggestion.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
subhash686: I see Dynamo persistence class has a private constructor too but I guess builder pattern instantiates the class there but MetricFactory probably does not get initiated the same way. 
Thanks for the suggestion, I used it as is.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
subhash686: Great idea. Added it.

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
subhash686: Modified per the suggestion.

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
subhash686: Thanks for catching this. I clearly overlooked that part. I caught ClassNotFoundException and `continue` is automatic there.

powertools-idempotency/powertools-idempotency-dynamodb/src/main/java/software/amazon/lambda/powertools/idempotency/persistence/dynamodb/DynamoDBPersistenceStore.java
subhash686: Modified per the suggestion.

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
subhash686: I will look into this. Did not change anything yet.

subhash686:
@phipag Thanks for your thorough review as always. I pushed the updates. 
I checked the 2 sonar issues and am inclined to accept those. Please let me know if you think otherwise.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
phipag: Yes, correct. The Builder in the Dynamo persistence store instantiates the class itself which is why the priming gets executed. The MetricsFactory however instantiates the provider which is a separate class.

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
phipag: Nice, thanks.

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
phipag: Awesome, thank you!

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
phipag: The input stream in the method below is a second input stream which is closed properly but I think this one needs proper closing as well. 

```java
try(InputStream is = conn.getInputStream();) {
                preloadClassesFromStream(is);
}
```

phipag:
> @phipag Thanks for your thorough review as always. I pushed the updates. I checked the 2 sonar issues and am inclined to accept those. Please let me know if you think otherwise.

Thanks very much for addressing my comments so quickly @subhash686. I resolved all comments which are fixed now. Please review the open comments.

Regarding Sonarcube, I accepted one issue. Let's address the second one in the next commit. We can extract the class loading into a method like `loadClassIfAvailable(className)`. 

I tested the code again after your updates in my AWS account. All Snapstart hooks run as expected now without errors.

powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
subhash686: Because InputStream is passed by reference and the try-with-resource handles the resource variables that are previously initialized as well, so the InputStream from line 45 must get closed properly. 

```java
private static void preloadClassesFromStream(InputStream is) {
        try (is; 
```

If you think its confusing doing it this way, I will remove `is` from this try block and move it up to to the caller. I dont think its right to include it twice.

phipag: You are right, I see the `is;` part of the `try` now. Let's keep it as is then.

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
subhash686: I have a `classessloaded.txt` file already for tests but I did not test whether valid classess are loaded specially when the file is a mixed bag of valid and invalid classes. Again thanks for catching that bug. 

Its a little tricky to test the static methods here but I got a little creative with `volatile` to not complicate tests with modifying thread context and cleanup. 

Finally, I checked that my new assertion fails with old code where if one class is not found the rest of the classes are skipped. And it works after fixing that code.

Let me know know if you see any more opportunities here.

subhash686:
> > @phipag Thanks for your thorough review as always. I pushed the updates. I checked the 2 sonar issues and am inclined to accept those. Please let me know if you think otherwise.
> 
> Thanks very much for addressing my comments so quickly @subhash686. I resolved all comments which are fixed now. Please review the open comments.
> 
> Regarding Sonarcube, I accepted one issue. Let's address the second one in the next commit. We can extract the class loading into a method like `loadClassIfAvailable(className)`.
> 
> I tested the code again after your updates in my AWS account. All Snapstart hooks run as expected now without errors.

Fixed this Sonarqube finding.

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
```diff
@@ -0,0 +1,34 @@
+package software.amazon.lambda.powertools.common.internal;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class ClassPreLoaderTest {
+
+    // Making this volatile so the Thread Context doesn't need any special handling
+    static volatile boolean dummyClassLoaded = false;
```
phipag, line 10: Out of interest: Can you explain what happens if don't put volatile here? I do not 100% follow the comment.

powertools-logging/powertools-logging-logback/src/test/java/software/amazon/lambda/powertools/logging/internal/LambdaJsonEncoderTest.java
```diff
@@ -368,7 +368,7 @@ void shouldLogThreadInfo() {
         String result = new String(encoded, StandardCharsets.UTF_8);
 
         // THEN
-        assertThat(result).contains("\"thread\":\"main\",\"thread_id\":1,\"thread_priority\":5");
+        assertThat(result).contains("\"thread\":\"main\",\"thread_id\":"+ Thread.currentThread().getId() +",\"thread_priority\":5");
```
phipag, line 371: Why was this change necessary? I think the change makes sense, just wondering if this test failed for you?

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
phipag: Wow, this unit test is very creative. I didn't think that works – but apparently it does. Thanks, resolving this comment. I added a small question here https://github.com/aws-powertools/powertools-lambda-java/pull/1861#discussion_r2245583315

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
phipag: FYI. Tests pass (https://github.com/aws-powertools/powertools-lambda-java/actions/runs/16651818872/job/47126553427?pr=1861).

subhash686: I will not be able to modify the dummyClassLoaded as if its a local variable. I am setting it to false and then checking ClassPreLoader::preLoadClasses turns it to true if it loaded successfully. 
I did not witness in handful runs but theoretically there is a possibility of flakiness in tests as the variable change from classloader may not available immediately. 
Alternate option is to set and reset `Thread.currentThread().setContextClassLoader`.

At least thats my understanding, I ran the tests almost 7 times and it worked every time with or without volatile.

powertools-logging/powertools-logging-logback/src/test/java/software/amazon/lambda/powertools/logging/internal/LambdaJsonEncoderTest.java
subhash686: Test is failing locally as my thread id is 3 but the test is expecting 1.

powertools-common/src/test/java/software/amazon/lambda/powertools/common/internal/ClassPreLoaderTest.java
phipag: I think I understand now:
1. Test initializes -> `dummyClassLoaded = false` is declared and initialized as `false`
2. ClassPreLoader runs
3. ClassPreLoader loads `DummyClass`
4. Static block triggers and sets `dummyClassLoaded = true` on the `ClassPreLoaderTest` class

It needs to be volatile such that the junit thread can immediatly assert the updated value based on the classloading of `DummyClass` executing the static block. 

Let me know if this makes sense.

subhash686: You've rightly put it!

phipag: Cool. Just saw your comment. I agree, let's keep `volatile`. Just wanted to understand better.

powertools-logging/powertools-logging-logback/src/test/java/software/amazon/lambda/powertools/logging/internal/LambdaJsonEncoderTest.java
phipag: Cool. Thanks for catching this!

phipag (approved):
@subhash686, I think this PR is ready to be merged now. I would like to congratulate you on this major contribution. This is laying the foundation for adding different CRaC priming techniques across this project. Also many thanks for documenting this process in the `Priming.md` doc. 🚀

In the following days, I will create follow-up issues on the items we discussed:
- Sub-issues for each Powertools utility referencing your Priming documentation
- An issue for automating the maintenance of the `classesloaded.txt` file through a GitHub workflow (implementation details to be defined).

subhash686:
@phipag Thank you for you support throughout. I had a great experience working on this feature. 
I will continue to look into Github workflow implementation ideas.

Issue #1864, opened 2025-06-06
Provide custom Executor Service for Parallel Batch Processing capabilities


visheshruparelia:
In this feature here: https://docs.powertools.aws.dev/lambda/java/preview/utilities/batch/#parallel-processing
It would be nice to have an option of providing my own ExecutorService instead of relying on the framework using its own thread pool. 
One way how a user can provide this is something like this: 
```
handler = new BatchMessageHandlerBuilder()
                .withSqsBatchHandler()
                .withExecutor(Executors.newFixedThreadPool(2)) // Can this be done?
                .buildWithMessageHandler(this::processMessage, Product.class);
```
A couple of reasons (not exhaustive, ofcourse) why this is useful could be: 
- User may want to have control over the concurrency
- User may want to control the names of the threads

PS - I am assuming parallel processing feature takes care of propagating the lambda context to all the threads via `ThreadContext/MDC` if used in conjuction with Powertools-logging. Thus, I can basically provide my own executor service and the framework will take care of propagating context to the threads I provided and also processing my sqs message in parallel. 

Let me know if this is possible to achieve.

phipag:
Hey @visheshruparelia,

thanks for opening this issue. I triaged this with the team and we decided to add this feature to the batch processing utility.

Would you like to help with the implementation?

visheshruparelia:
Hi @phipag thank you. Yes I don't mind.

phipag:
That is awesome to hear @visheshruparelia. Let's go with your proposal of supporting passing a custom executor but make a slight change in where we pass it. 

Instead of passing it in the Builder, let's create a method overload for the `handler.processBatchInParallel(sqsEvent, context);` method. Like this:

```java
// Optional third argument to pass a custom executor
handler.processBatchInParallel(sqsEvent, context, Executors.newFixedThreadPool(2));
```

The reason is that the Builder is for all Batch processing configurations but thread pools are specific to parallel processing only. This way we reduce the public surface area of custom executor configurations. 

Here are some high-level steps you can follow:
- Implement [the method overload](https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-batch/src/main/java/software/amazon/lambda/powertools/batch/handler/BatchMessageHandler.java#L52) and update the `processBatchInParallel` method in the implementations
[...]

PR #1900, opened 2025-06-18
feat: add support for batch execution in parallel with custom Executor

visheshruparelia:
#1864

## Description of changes:

Add a new method in the BatchHandler class which allows users to pass their custom Executor for processing messages in parallel.

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/docs/utilities/batch.md
+++ b/docs/utilities/batch.md
@@ -512,7 +512,7 @@ used with SQS FIFO. In that case, an `UnsupportedOperationException` is thrown.
     ```
 === "Example with SQS (using custom executor)"
 
-    ```java hl_lines="13"
+    ```java hl_lines="4 10 15"
     public class SqsBatchHandler implements RequestHandler<SQSEvent, SQSBatchResponse> {
 
         private final BatchMessageHandler<SQSEvent, SQSBatchResponse> handler;


powertools-batch/src/main/java/software/amazon/lambda/powertools/batch/internal/MultiThreadMDC.java
```diff
@@ -44,4 +44,11 @@ public void copyMDCToThread(String thread) {
             mdcAwareThreads.add(thread);
         }
     }
+
+    public void removeThread(String thread) {
```
visheshruparelia, line 48: I think this will be helpful if the same invocation of `processBatchInParallel()` ends up re-using the thread.

visheshruparelia:
@phipag sonar is failing due to duplication in Kinesis and DynamoDB handlers. The same duplication exists for the already existing method as well. As a result, I followed the same pattern for the new method too. If you want, we can move the duplicated block for both (older and newer) methods to a common static utlilty method but IMO this duplication is ok. Let me know.

dreamorosi:
@phipag could you please unblock this during this iteration? I think there's some pending feedback we need to give to the contributor. Thanks!

examples/powertools-examples-batch/src/main/java/org/demo/batch/kinesis/KinesisBatchHandlerParallel.java
```diff
@@ -0,0 +1,39 @@
...
+                .buildWithMessageHandler(this::processMessage, Product.class);
+        executor = Executors.newFixedThreadPool(2);
+    }
+
+    @Override
+    public StreamsEventResponse handleRequest(KinesisEvent kinesisEvent, Context context) {
+        return handler.processBatchInParallel(kinesisEvent, context, executor);
+    }
+
+    private void processMessage(Product p, Context c) {
```
phipag: You can remove `Context c` here please. There is an overload for a message handler without context in `AbstractBatchMessageHandlerBuilder.java`

```java
    /**
     * Builds a BatchMessageHandler that can be used to process batches, given
     * a user-defined handler to process each item in the batch. This variant
     * takes a function that consumes the deserialized body of the given message
     * If deserialization fails, it will be treated as
     * failure of the processing of that item in the batch.
     * Note:  If you don't need the Lambda context, use the variant of this function
     * that does not require it.
     *
     * @param handler Processes the deserialized body of the message
     * @return A BatchMessageHandler for processing the batch
     */
    public <M> BatchMessageHandler<E, R> buildWithMessageHandler(Consumer<M> handler, Class<M> messageClass) {
        return buildWithMessageHandler((f, c) -> handler.accept(f), messageClass);
    }
```

examples/powertools-examples-batch/src/main/java/org/demo/batch/dynamo/DynamoDBStreamBatchHandlerParallel.java
```diff
@@ -0,0 +1,37 @@
...
+                .buildWithRawMessageHandler(this::processMessage);
+        executor = Executors.newFixedThreadPool(2);
+    }
+
+    @Override
+    public StreamsEventResponse handleRequest(DynamodbEvent ddbEvent, Context context) {
+        return handler.processBatchInParallel(ddbEvent, context, executor);
+    }
+
+    private void processMessage(DynamodbEvent.DynamodbStreamRecord dynamodbStreamRecord, Context context) {
```
phipag: You can remove `Context c` here please. There is an overload for a message handler without context in `AbstractBatchMessageHandlerBuilder.java`

```java
    /**
     * Builds a BatchMessageHandler that can be used to process batches, given
     * a user-defined handler to process each item in the batch. This variant
     * takes a function that consumes a raw message and the Lambda context. This
     * is useful for handlers that need access to the entire message object, not
     * just the deserialized contents of the body.
     *
     * @param handler Takes a raw message - the underlying AWS Events Library event - to process.
     *                For instance for SQS this would be an SQSMessage.
     * @return A BatchMessageHandler for processing the batch
     */
    public BatchMessageHandler<E, R> buildWithRawMessageHandler(Consumer<T> handler) {
        return buildWithRawMessageHandler((f, c) -> handler.accept(f));
    }
```

powertools-batch/src/main/java/software/amazon/lambda/powertools/batch/handler/KinesisStreamsBatchMessageHandler.java
```diff
@@ -77,7 +81,9 @@ public StreamsEventResponse processBatchInParallel(KinesisEvent event, Context c
                 .parallelStream() // Parallel processing
                 .map(eventRecord -> {
                     multiThreadMDC.copyMDCToThread(Thread.currentThread().getName());
-                    return processBatchItem(eventRecord, context);
+                    Optional<StreamsEventResponse.BatchItemFailure> failureOpt = processBatchItem(eventRecord, context);
+                    multiThreadMDC.removeThread(Thread.currentThread().getName());
```
phipag, line 85: Good catch!!

powertools-batch/src/main/java/software/amazon/lambda/powertools/batch/internal/MultiThreadMDC.java
phipag: Yes, this is good. I believe it was a small memory leak because we never removed the thread either.

phipag (changes requested):

dreamorosi:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=aws-powertools_powertools-lambda-java&pullRequest=1900) **Quality Gate failed**
> Failed conditions  [9.8% Duplication on New Code](https://sonarcloud.io/component_measures?id=aws-powertools_powertools-lambda-java&pullRequest=1900&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=aws-powertools_powertools-lambda-java&pullRequest=1900)

How come did the code duplication increase so much? Is this old code that is now being scanned, or new code? Can we do anything to improve it?

phipag:
Hey @visheshruparelia, I finally reviewed your PR and it works great. I just tested it end-to-end using the examples you also added. Thanks for this! 🚀 

Can you please have a look at the pmd_analyse comments in the diff and address them? I added comments to some but there are more such as:

> A catch statement should never catch throwable since it includes errors.
> Catching Throwable errors is not recommended since its scope is very broad. It includes runtime issues such as
OutOfMemoryError that should be exposed and managed separately.
> AvoidCatchingThrowable (Priority: 1, Ruleset: Error Prone)
https://docs.pmd-code.org/snapshot/pmd_rules_java_errorprone.html#avoidcatchingthrowable

I think we can just catch `Exception` there.

Regarding the code duplication reported by Sonarcube, I think it is okay in this case since it is actually easier to read for developers than creating more abstractions in addition to the existing ones.

One last thing is the documentation page [here](https://github.com/aws-powertools/powertools-lambda-java/blob/1fa11c138529ee7d49f6d24415a5de6223330aa2/docs/utilities/batch.md). Do you mind documenting the new feature that you implemented there? You could add a second example showing how to pass a custom executor to this code snippet as a second tab:

image

phipag:
> > ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=aws-powertools_powertools-lambda-java&pullRequest=1900) **Quality Gate failed**
> > Failed conditions  [9.8% Duplication on New Code](https://sonarcloud.io/component_measures?id=aws-powertools_powertools-lambda-java&pullRequest=1900&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> > [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=aws-powertools_powertools-lambda-java&pullRequest=1900)
> 
> How come did the code duplication increase so much? Is this old code that is now being scanned, or new code? Can we do anything to improve it?

The code duplication was there before and triggered again as a finding in this PR due to implementing the new method overload for `processBatchInParallel`:


```java
    @Override
    public StreamsEventResponse processBatchInParallel(DynamodbEvent event, Context context, Executor executor) {
        MultiThreadMDC multiThreadMDC = new MultiThreadMDC();

        List<StreamsEventResponse.BatchItemFailure> batchItemFailures = new ArrayList<>();
        List<CompletableFuture<Void>> futures = event.getRecords().stream()
                .map(eventRecord -> CompletableFuture.runAsync(() -> {
                    multiThreadMDC.copyMDCToThread(Thread.currentThread().getName());
                    Optional<StreamsEventResponse.BatchItemFailure> failureOpt = processBatchItem(eventRecord, context);
                    failureOpt.ifPresent(batchItemFailures::add);
                    multiThreadMDC.removeThread(Thread.currentThread().getName());
                }, executor))
                .collect(Collectors.toList());
        futures.forEach(CompletableFuture::join);
        return StreamsEventResponse.builder().withBatchItemFailures(batchItemFailures).build();
    }
```

This code snippet is essentially the same for DynamoDB and Kinesis (code duplication reported here) – but it is different for SQS because we need to check if we are dealing with a FIFO queue. It would require some larger refactoring if we wanted to remove the code duplication completely. 

My personal opinion is, that we should favor the current code duplication because:
- It is more readable (we don't have to jump between Abstract class **and** interface implementations)
- It is more flexible if we need to make adjustments for details of a specific event source (e.g. SQS FIFO)
- It is easier for developers to add a new implementation for a new event source. Just implement the `BatchMessageHandler` interface. I don't need to refactor again some abstracted logic if the new event source has special requirements.

If we think that we should refactor though, I am also okay with that. Let me know what you folks think @dreamorosi and @visheshruparelia .

dreamorosi:
> This code snippet is essentially the same for DynamoDB and Kinesis (code duplication reported here) – but it is different for SQS because we need to check if we are dealing with a FIFO queue. It would require some larger refactoring if we wanted to remove the code duplication completely.
> 
> My personal opinion is, that we should favor the current code duplication because:


Sounds sensible, thanks for explaining.

visheshruparelia:
Hi @phipag, thank you for the review. 

>Can you please have a look at the pmd_analyse comments in the diff and address them?
Most are on the code which is untouched in this PR. Would you prefer combine unrelated refactoring + the main code changes in the same PR?

examples/powertools-examples-batch/src/main/java/org/demo/batch/dynamo/DynamoDBStreamBatchHandlerParallel.java
visheshruparelia: The older examples in the package follow the same pattern as the new one. Do you want those changed as well?

phipag: Thanks, this would be awesome if you can update those as well. Ideally, we have no Sonar or pmd_analyze findings before merging the PR. With the exception of the code duplication finding.

visheshruparelia: 850c15542ddd1ff8026d2020e4263bebb61dced3

examples/powertools-examples-batch/src/main/java/org/demo/batch/kinesis/KinesisBatchHandlerParallel.java
visheshruparelia: 850c15542ddd1ff8026d2020e4263bebb61dced3

visheshruparelia:
> Hey @visheshruparelia, I finally reviewed your PR and it works great. I just tested it end-to-end using the examples you also added. Thanks for this! 🚀
> 
> Can you please have a look at the pmd_analyse comments in the diff and address them? I added comments to some but there are more such as:
> 
> > A catch statement should never catch throwable since it includes errors.
> > Catching Throwable errors is not recommended since its scope is very broad. It includes runtime issues such as
> > OutOfMemoryError that should be exposed and managed separately.
> > AvoidCatchingThrowable (Priority: 1, Ruleset: Error Prone)
> > https://docs.pmd-code.org/snapshot/pmd_rules_java_errorprone.html#avoidcatchingthrowable
> 
> I think we can just catch `Exception` there.
> 
> Regarding the code duplication reported by Sonarcube, I think it is okay in this case since it is actually easier to read for developers than creating more abstractions in addition to the existing ones.
> 
> One last thing is the documentation page [here](https://github.com/aws-powertools/powertools-lambda-java/blob/1fa11c138529ee7d49f6d24415a5de6223330aa2/docs/utilities/batch.md). Do you mind documenting the new feature that you implemented there? You could add a second example showing how to pass a custom executor to this code snippet as a second tab:
> 
> image

4d0cd06, 850c15542ddd1ff8026d2020e4263bebb61dced3

docs/utilities/batch.md
```diff
@@ -510,6 +510,31 @@ used with SQS FIFO. In that case, an `UnsupportedOperationException` is thrown.
         }
     }
     ```
+=== "Example with SQS (using custom executor)"
+
+    ```java hl_lines="13"
```
phipag:
```suggestion
    ```java hl_lines="4 10 15"
```

phipag (changes requested):

phipag:
Thanks very much for your update @visheshruparelia! Can you just commit the suggestion that I made to highlight the correct lines in your new doc example? After this, we are ready to merge 🚀

docs/utilities/batch.md
phipag: image

visheshruparelia:
@phipag done

Issue #1901, opened 2025-06-19
Feature: Support Confluent producers in new Kafka utility


phipag:
**Only for protobuf:**

According to the wire format of the Confluent serializers message-indices are injected as additional bytes after magic bytes and schema id:

*magic-byte, schema-id, **message-indexes**, protobuf-payload*

We need to strip these message-indexes before attempting to parse the data using the plain Google protobuf library if they are present.

See: https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#wire-format

PR #1902, opened 2025-06-19
fix(kafka): Add support for confluent message indices.

phipag:
**Issue #, if available:** https://github.com/aws-powertools/powertools-lambda-java/issues/1901

## Description of changes:

This is an import fix to support customers using Confluent schema integration on the producer side of Protobuf messages. This only applies for Protobuf and **not** for Avro. 

According to the wire format of the Confluent serializers message-indices are injected as additional bytes after magic bytes and schema id:

*magic-byte, schema-id, **message-indexes**, protobuf-payload*

We need to strip these message-indexes before attempting to parse the data using the plain Google protobuf library. This is what this PR is doing:

1. Attempt to parse using plain protobuf
2. If it fails
    1. Attempt to strip message-indexes and parse again
3. If it still fails we have some corrupted data or a schema mismatch and raise an exception

Added unit tests covering 100% of branches.

image

**Checklist**

 

* [x] Update tests
* [x] Update docs

### DIFF
--- a/powertools-kafka/src/main/java/software/amazon/lambda/powertools/kafka/serializers/KafkaProtobufDeserializer.java
+++ b/powertools-kafka/src/main/java/software/amazon/lambda/powertools/kafka/serializers/KafkaProtobufDeserializer.java
@@ -13,13 +13,26 @@
 package software.amazon.lambda.powertools.kafka.serializers;
 
 import java.io.IOException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.protobuf.CodedInputStream;
 import com.google.protobuf.Message;
 import com.google.protobuf.Parser;
 
 /**
  * Deserializer for Kafka records using Protocol Buffers format.
+ * Supports both standard protobuf serialization and Confluent Schema Registry serialization using messages indices.
+ * 
+ * For Confluent-serialized data, assumes the magic byte and schema ID have already been stripped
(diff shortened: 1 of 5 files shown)


leandrodamascena (approved):

powertools-kafka/src/main/java/software/amazon/lambda/powertools/kafka/serializers/KafkaProtobufDeserializer.java
```diff
@@ -40,4 +55,45 @@ protected <T> T deserializeObject(byte[] data, Class<T> type) throws IOException
...
+                // Simple case: Single 0 byte means first message type
+                LOGGER.debug("Found simple message-index case (single 0), parsing remaining data as protobuf");
+                return parser.parseFrom(codedInputStream);
+            } else {
+                // Complex case: firstValue is the length of the message index array
+                LOGGER.debug("Found complex message-index case with array length: {}, skipping {} message index values",
+                        firstValue, firstValue);
+                for (int i = 0; i < firstValue; i++) {
+                    codedInputStream.readUInt32(); // Skip each message index value
+                }
```
duckinghipster, line 89: This will clip multiple bytes which might be part of the actual data as well. We have seen cases for confluent serialized proto messages where first byte is not the length but index of the message

phipag: Thanks. I will update to reading varint and also handle Glue bytes as an additional edge-case.

Issue #1803, opened 2025-03-13
Maintenance: Build and test GraalVM native images in GH Workflows for v2


phipag:
### Summary

After introducing GraalVM support for the Core Utilities (https://github.com/aws-powertools/powertools-lambda-java/issues/764), we need to add an additional step to the GitHub build workflows (on v2 branch) to build and test the native image in addition to JVM.

### Why is this needed?

This is needed to avoid regressions in GraalVM compatibility in the future and detect unsupported or breaking changes regarding native compilation before merging / releasing them. There is a related issue for running end-to-end tests to test the native images at runtime https://github.com/aws-powertools/powertools-lambda-java/issues/1805.

### Which area does this relate to?

Governance, GraalVM

### Solution

The expected behavior is that the workflow fails if either the GraalVM Reachability Metadata generation fails or if the unit tests against the native image fail. Otherwise, the workflow should succeed.

It should also be possible to skip enable/disable native compilation and testing for specific submodules. This can be done by using the `-pl` argument in Maven. For example, if I would like to generate GRM files and run unit tests against the native image for all modules except for `powertools-common` I can run this command to exclude `powertools-idempotency`:

```shell
mvn -Pgenerate-graalvm-files clean test -pl "-powertools-idempotency" && mvn -Pgraalvm-native clean test -pl "-powertools-idempotency"
```

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda (Java) languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

phipag:
Related PR for E2E tests: https://github.com/aws-powertools/powertools-lambda-java/pull/1945.

phipag:
We had to introduce Mockito `5.19.1-SNAPSHOT` as part of the PR. This issue tracks moving a stable release again as soon as it is released: https://github.com/aws-powertools/powertools-lambda-java/issues/2079

PR #1945, opened 2025-07-18
chore(ci): Add GraalVM E2E tests and GH workflows

phipag:
## Summary

This PR introduces E2E tests for testing against native images deployed on `AL2023.provided` runtime compiled using GraalVM.

It is important to have these tests to cover regressions of future code changes. For example, if the GraalVM metadata files were not updated inline with new code changes. We want to protect customers from releasing changes that break GraalVM compatibility.

Tests pass ✅
- https://github.com/aws-powertools/powertools-lambda-java/actions/runs/16373230793

### Changes

The GraalVM E2E tests are triggered by a custom `e2e-graal` maven profile which sets a System Property `graalvm.enabled = true`. This will be read within the E2E infrastructure logic to decide whether to deploy a native image on the Lambda provided runtime or not. The E2E test assertions themselves stay untouched because we expect the exact same output for native images like for JVM based Lambda handlers. We can include more utilities in GraalVM E2E tests as we grow the GraalVM support across more Powertools utilities by updating this maven profile configuration.

```xml
 <configuration>
      <skipAfterFailureCount>1</skipAfterFailureCount>
      <includes>
          <!-- Add additional E2E tests here when adding new GraalVM support to utilities. -->
          <include>**/MetricsE2ET.java</include>
          <include>**/LoggingE2ET.java</include>
          <include>**/ParametersE2ET.java</include>
      </includes>
      <excludes>
          <exclude>**/TracingE2ET.java</exclude>
      </excludes>
      <systemPropertyVariables>
          <graalvm.enabled>true</graalvm.enabled>
      </systemPropertyVariables>
  </configuration>
```

- Update GitHub action workflow to include `e2e-graal` job running the same Java matrix and same workflows like `e2e`
- Update E2E test handlers to include GraalVM metadata files required to run as native image (this will be same experience like an end user would do when deploying a Lambda handler)
- Update Infrastructure configuration to be compatible with GraalVM
    - Introduced new `DockerConfiguration.java` to hold configuration for CDK Bundling within Docker. Required for x86 cross-compilation of native image.

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/1803

---

### DIFF
--- a/.github/workflows/check-e2e.yml
+++ b/.github/workflows/check-e2e.yml
@@ -47,6 +47,7 @@ jobs:
       id-token: write
     environment: E2E
     strategy:
+      fail-fast: false
       max-parallel: 3
       matrix:
         java:
@@ -79,6 +80,7 @@ jobs:
       id-token: write
     environment: E2E
     strategy:
+      fail-fast: false
       max-parallel: 3
       matrix:
         java:


phipag:
I don't understand why the osv-scanner fails since this vulnerability was not introduced by this PR. According to the docs it should only scan the PR diff. https://google.github.io/osv-scanner/github-action/#scan-on-pull-request

.github/workflows/check-e2e.yml
```diff
@@ -70,4 +70,36 @@ jobs:
...
+    permissions:
+      id-token: write
+    environment: E2E
+    strategy:
+      max-parallel: 3
+      matrix:
+        java:
+          - 11
+          - 17
+          - 21
```
leandrodamascena: In this case, you might have issues with Java 11 but not Java 17, for example, so I think it makes sense to add `fail-fast: false` to this strategy so you can run e2e and not stop if a specific version fails. This makes debugging easier and you don't waste time wondering if the problem occurs in all versions.

powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/MetricsE2ET.java
```diff
@@ -84,42 +84,40 @@ public void test_recordMetrics() {
         // THEN
         MetricsFetcher metricsFetcher = new MetricsFetcher();
         List<Double> coldStart = metricsFetcher.fetchMetrics(invocationResult.getStart(), invocationResult.getEnd(), 60,
```
leandrodamascena, line 86: Since you are setting [retry strategy](https://github.com/aws-powertools/powertools-lambda-java/blob/phipag/graalvm-e2e/powertools-e2e-tests/src/test/java/software/amazon/lambda/powertools/testutils/metrics/MetricsFetcher.java#L104) in the SDK session, it's fine to not have retry here. Nice one.

powertools-e2e-tests/src/test/resources/docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Use the official AWS SAM base image for Java 21
+FROM public.ecr.aws/sam/build-java21:latest
```
leandrodamascena: Please change this to use the sha image and then dependabot can bump it when we have a new version of this image. Scorecard rules doesn't allow using latest because you don't have control when it can get updated and probably will broke your tests.

```sh
docker manifest inspect public.ecr.aws/sam/build-java21:latest

{
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 3903,
         "digest": "sha256:a5554d68374e19450c6c88448516ac95a9acedc779f318040f5c230134b4e461",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      }

```

phipag: https://github.com/aws-powertools/powertools-lambda-java/pull/1945/commits/70c7c0ecb1306f15b3e841e97620ee3cde17dc53

.github/workflows/check-e2e.yml
phipag: https://github.com/aws-powertools/powertools-lambda-java/pull/1945/commits/b5682578bc81afa8fa2aaf9482264ad2a03ceaff

phipag:
Thanks for your feedback @leandrodamascena, I addressed your comments.

Issue #1981, opened 2025-07-28
Bug: Regression error in v2 AppConfig parameters


chrisclayson:
### Expected Behaviour

If AppConfig returns an empty response (indicating that there have been no changes to the config since the last call to `GetLatestConfiguration`), then the provider should continue to return the cached version of the config.

### Current Behaviour

As per the [issue](https://github.com/aws-powertools/powertools-lambda-java/issues/1672) raised on v1 of the app, v2 is not correctly handling the empty response.

If an empty response is retrieved from AppConfig then `null` is returned which causes any transformer or, ultimately, the calling code to fail.

It seems the [PR](https://github.com/aws-powertools/powertools-lambda-java/pull/1673) that was raised to fix the problem in v1 was not migrated to v2.

### Code snippet

```java
var provider = AppConfigProvider.builder()
                        .withClient(appConfigDataClient)
                        .withEnvironment(environmentVariables.getAppConfigEnvironment())
                        .withApplication(environmentVariables.getAppConfigApplication())
                        .build();

...

// Works fine on first call, but subsequent calls return `null`
var config = provider
    provider
        .withMaxAge(getConfigurationCacheTtl(), ChronoUnit.SECONDS)
        .withTransformation(ConfigTransformer.class)
        .get("test", Config.class);
```

### Possible Solution

See PR #1673

### Steps to Reproduce

- Setup a freeform configuration profile in AppConfig with some value (we used a JSON object `{ "version": 1 }`)
- Retrieve from a lambda using PowerTools parameters:

```java
          var config = ParamManager.getAppConfigProvider("test", "test")
                .withMaxAge(30, ChronoUnit.SECONDS)
                .get("test");
```
- Fire continuous traffic at the lambda

The first retrieval, on any given runtime environment,  will work fine, but after the 30 second TTL the next retrieve from AppConfg returns an empty string, which is returned from the `get()` method rather than original (cached value).

### Powertools for AWS Lambda (Java) version

2.2.0 (latest)

### AWS Lambda function runtime

Java 17

### Debugging logs

```java

```

phipag:
Hey @chrisclayson,

thanks for reporting this bug. You are right, it appears the code slipped migration when refactoring the parameters module into sub-modules. I see the same problem when running in my AWS account.

We need to re-introduce the changes from PR https://github.com/aws-powertools/powertools-lambda-java/pull/1673 again in this code snippet:

[...]

PR #1982, opened 2025-07-29
fix(parameters): Correctly check for empty values in AppConfig Parameters Provider.

phipag:
## Summary

This PR fixes an issue where the AppConfig Provider wrongly returns null. AppConfig returns an empty byte array on `GetLatestConfigurationChange` if there was no change since the session was established. It is not sufficient to perform a null check but rather we should check if the byte array is of length zero.

This PR is similar to https://github.com/aws-powertools/powertools-lambda-java/pull/1673 against v1 but was not migrated correctly to v2.

Related docs: https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-code-samples-using-API-read-configuration.html

Also resolves existing Sonarcube linting errors where possible.

### Changes

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/1981

---


leandrodamascena (approved):
Approved!

Issue #2021, opened 2025-08-07
Docs: Create example for using Bazel build system


phipag:
### What were you searching in the docs?

For customers who do not use Maven or Gradle it can be useful to have a code example on how to use the different utilities with the [Bazel build system](https://bazel.build/). 

Especially, the configuration of AspectJ compile-time weaving is not straightforward and can be an adoption blocker for customer who wish to use Bazel.

### How can we improve?

Create a fully working code example using Bazel instead of Maven/Gradle. The example should contain a way to deploy to AWS, for example, using SAM.

### Acknowledgment

- [x] I understand the final update might be different from my proposed suggestion, or refused.

PR #2022, opened 2025-08-07
docs(examples): Add Bazel example for core utilities

phipag:
## Summary

This PR adds a new code example `sam-bazel` showcasing how to use the Core Utilities including AspectJ compile-time weaving using Bazel.

### Changes

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/2021

---

### DIFF
--- a/examples/powertools-examples-core-utilities/sam-bazel/.bazelrc
+++ b/examples/powertools-examples-core-utilities/sam-bazel/.bazelrc
@@ -1,0 +1,27 @@
+# Java compilation settings
+build --java_language_version=11
+build --java_runtime_version=11
+
+# Enable worker strategy for better performance
+build --strategy=Javac=worker
+build --worker_sandboxing
+
+# Modern Java build optimizations
+build --experimental_strict_java_deps=off
+build --incompatible_java_common_parameters=false
+
+# Performance optimizations
+build --experimental_reuse_sandbox_directories
+build --worker_multiplex
+
+# Disk cache for better performance
+build --disk_cache=~/.cache/bazel-disk-cache
+
+# Remote cache configuration (optional)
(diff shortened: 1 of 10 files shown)


phipag:
At the moment we have this section: https://docs.powertools.aws.dev/lambda/java/latest/core/logging/#getting-started. Which references the examples folder on GitHub. Although this is only the case for the Logging utility. 

I don't think we should specifically highlight the Bazel example and not the other examples. One could argue why Bazel is more important than for example our Gradle or Kotlin examples. 

But I agree with your comment that the examples are not referenced in our documentation in a popular way. Do you have something in mind what we can do? Maybe on your homepage? We can add a box such as

> **Looking for examples?**
> Checkout the full list of LINK TO EXAMPLES including usage examples for Gradle, Bazel, and Kotlin.

dreamorosi:
Fair, then let's leave it as is for now.

Not in this PR, but in the near future we could consider introducing a similar structure as we have in TypeScript with a "Getting Started" collection of pages, for example [this one](https://docs.powertools.aws.dev/lambda/typescript/latest/getting-started/installation/) shows how to install different packages and [this](https://docs.powertools.aws.dev/lambda/typescript/latest/getting-started/usage-patterns/) shows different patterns, and [this one TypeScript settings](https://docs.powertools.aws.dev/lambda/typescript/latest/getting-started/typescript-settings/).

Obviously in the case of Java the content would be different, but this would probably be a good place to show these type of things, explain nuances around AspecJ, Crac, GraalVM, and similar things - in addition to utility-specific info on the respective pages.

dreamorosi (approved):
Thanks for working and adding this quickly!

phipag:
Thanks. I like these sections.

Issue #2095, opened 2025-08-28
Feature request: Log buffering support for Logger utility


phipag:
### Use case

Log buffering enables you to buffer logs for a specific request or invocation, and flush them automatically on error or manually as needed. This is useful when you want to reduce the number of log messages emitted while still having detailed logs when needed, such as when troubleshooting issues.

This feature is already available in other Powertools language implementations:
- [Python](https://docs.powertools.aws.dev/lambda/python/latest/core/logger/#buffering-logs)
- [TypeScript](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/#buffering-logs)
- [.NET](https://docs.powertools.aws.dev/lambda/dotnet/core/logging/#buffering-logs)

**PLEASE READ:**
- Log buffering documentation of the other languages
- Final specification from original RFC: https://github.com/aws-powertools/powertools-lambda-typescript/discussions/3410#discussioncomment-12097189

The remainder of this issue outlines a high-level overview of how this experience can look like in Java.

### Complete Example
```java
public class MyHandler implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponse> {
    private static final Logger logger = LoggerFactory.getLogger(MyHandler.class);
    
    public MyHandler() {
        // Programmatic configuration using Builder pattern
        PowertoolsLogging.setBufferConfig(BufferConfig.builder()
            .maxBytes(20480)  // Default is 20KB (20480 bytes)
            .bufferAtVerbosity(Level.DEBUG)  // Buffers DEBUG and TRACE
            .flushOnErrorLog(true)  // default true, auto-flushes buffer when error logs are emitted
            .compress(false)  // default false
            .build());
            
        // INIT logs will never be buffered due to no trace ID being present
        logger.debug("This is a debug message");  // This is NOT buffered
    }
    
    @Logging
    public APIGatewayProxyResponse handleRequest(APIGatewayProxyRequestEvent event, Context context) {
        // Assuming current log level is INFO
        logger.debug("This is a debug message");  // This is buffered
        logger.info("This is an info message");
        
        // your business logic here
        
        logger.error("This is an error message");  // This also flushes the buffer
        
        return new APIGatewayProxyResponse();
    }
}
```

### Annotation-based Configuration
```java
@Logging(bufferConfig = @BufferConfig(maxBytes = 20480, flushOnErrorLog = true))
public String handleRequest(APIGatewayProxyRequestEvent event, Context context) {
[...]

PR #2103, opened 2025-09-01
feat(logging): Log buffering support for Logj42 and Logback

phipag:
## Summary

Docs preview: https://dealn7fl31ram.cloudfront.net/core/logging/#buffering-logs

This PR introduces a log buffering feature to Powertools for AWS Lambda (Java). Log buffering is the mechanism of buffering lower level logs and only flush them (to STDOUT on Lambda) in case of error or manually if needed. It is a good way to reduce the noise caused by e.g. DEBUG logs and can be seen as a more elaborate way of "log sampling". 

While the other Powertools runtimes ([Python](https://docs.powertools.aws.dev/lambda/python/latest/core/logger/#buffering-logs), [TypeScript](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/#buffering-logs), [.NET](https://docs.powertools.aws.dev/lambda/dotnet/core/logging/#buffering-logs)) follow a programmatic approach to configuring log buffering, this PR uses the concept of "Appenders" in the corresponding logging backend (log4j2, logback). This means that log buffering is configured using the configuration files of the respective logging framework which is compatible with the way how structured JSON logging is already implemented. This also means that log buffering can be used independent of the `@Logging` annotation (except for automatic flushing on exceptions) and it is also compatible with any other Appender. The idea is to wrap an arbitrary existing Appender (e.g. a STDOUT JSON structured appender) with the `BufferingAppender` which will enable log buffering with zero code changes. Here is an example for log4j2:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Configuration>
    <Appenders>
        <Console name="JsonAppender" target="SYSTEM_OUT">
            <JsonTemplateLayout eventTemplateUri="classpath:LambdaJsonLayout.json" />
        </Console>
        <!-- New BufferingAppender wrapping existing JsonAppender emitting Lambda structured JSON logs -->
        <BufferingAppender name="BufferedAppender" bufferAtVerbosity="DEBUG" maxBytes="20480">
            <AppenderRef ref="JsonAppender" />
        </BufferingAppender>
    </Appenders>
    <Loggers>
        <Root level="debug">
            <!-- Reference the BufferedAppender wrapping the JsonAppender instead of the JsonAppender directly -->
            <AppenderRef ref="BufferedAppender" />
        </Root>
    </Loggers>
</Configuration>
```

Note: This XML based configuration supersedes the original design proposal of `@BufferingConfig` made in the feature request issue (see below). 

### Changes

- Implement log buffering feature
- Extend E2E tests to test for both log4j2 and logback implementations (we only had log4j2 before)
- Update examples to use log buffering feature
- Add section in documentation about log buffering
- Make buffering compatible with GraalVM

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/2095

---


phipag:
#### TODO
- ~~Write developer documentation~~ https://dealn7fl31ram.cloudfront.net/core/logging/#buffering-logs
- ~~Address Sonar findings~~ ✅
- ~~Address PMD findings~~ ✅
- ~~Understand better if trace id is the best key for the Log buffering `KeyBuffer`~~ ✅
    - Yes, it is unique for each Lambda execution within `handleRequest`.
- ~~Revert original examples but add log buffering to them~~ ✅

docs/core/logging.md
```diff
@@ -787,7 +787,377 @@ with `logError` param or via `POWERTOOLS_LOGGER_LOG_ERROR` env var.
...
+    Lambda->>Client: Raise uncaught exception
+```
+<i>Flushing buffer when an uncaught exception happens</i>
+</center>
+
+#### Buffering FAQs
+
+1. **Does the buffer persist across Lambda invocations?** No, each Lambda invocation has its own buffer. The buffer is initialized when the Lambda function is invoked and is cleared after the function execution completes or when flushed manually.
+2. **Are my logs buffered during cold starts (INIT phase)?** No, we never buffer logs during cold starts. This is because we want to ensure that logs emitted during this phase are always available for debugging and monitoring purposes. The buffer is only used during the execution of the Lambda function.
+3. **How can I prevent log buffering from consuming excessive memory?** You can limit the size of the buffer by setting the `maxBytes` option in the `BufferingAppender` configuration. This will ensure that the buffer does not grow indefinitely and consume excessive memory.
```
svozza:
```suggestion
3. **How can I prevent log buffering from consuming excessive memory?** You can limit the size of the buffer by setting the `maxBytes` option in the `BufferingAppender` configuration. This will ensure that the buffer does not grow indefinitely.
```

docs/core/logging.md
```diff
@@ -787,7 +787,377 @@ with `logError` param or via `POWERTOOLS_LOGGER_LOG_ERROR` env var.
...
+4. **What happens if the log buffer reaches its maximum size?** Older logs are removed from the buffer to make room for new logs. This means that if the buffer is full, you may lose some logs if they are not flushed before the buffer reaches its maximum size. When this happens, we emit a warning when flushing the buffer to indicate that some logs have been dropped.
+5. **How is the log size of a log line calculated?** The log size is calculated based on the size of the log line in bytes. This includes the size of the log message, any exception (if present), the log line location, additional keys, and the timestamp.
+6. **What timestamp is used when I flush the logs?** The timestamp preserves the original time when the log record was created. If you create a log record at 11:00:10 and flush it at 11:00:25, the log line will retain its original timestamp of 11:00:10.
```
svozza:
```suggestion
6. **What timestamp is used when I flush the logs?** The timestamp is the original time when the log record was created. If you create a log record at 11:00:10 and flush it at 11:00:25, the log line will retain its original timestamp of 11:00:10.
```

dreamorosi:
Hi @svozza, did you by chance also review the implementation or just the docs?

svozza:
Yes, it's a big diff so I only really focused on the implementation's Java classes. The only thing I noticed was that `logback/BufferingAppender.java` and `log4j/BufferingAppender.java` seem to share quite a bit of common logic but I decided it probably isn't worth try to extract it.

dreamorosi:
@phipag comments on @svozza's feedback above?

If you think it's not necessary to extract I'm ok to merge as is.

svozza (approved):

phipag:
Thanks @svozza and @dreamorosi for the feedback. It is a good point and I thought about this as well. Some of the code flow is shared but there are notable differences in the typing and behavior of the log events / Appenders in these frameworks. 

The main candidate for abstraction would be the [append()](https://github.com/aws-powertools/powertools-lambda-java/pull/2103/files#diff-4f413edd96101704da43ded58fb658cadda1e7d10250b22264ad6c128cd05da8R107-R127) method. However, you can see that log4j2 requires to create a immutable copy of the log event first while logback doesn't. Also the log level comparison for the buffering decision is different due to the `Level` type being different.

Another argument I fought in my mind was whether the `LambdaHandlerProcessor.getXrayTraceId()` calls should be abstracted. I decided against this as well because this logic already comes from the `powertools-common` module. If we change the way how individual requests are identified in a Lambda execution we will change it in `powertools-common` anyway so that it propagates to all consumers of this logic. 

Historically, this is also the reason why the JSON structured logging was implemented independently for both log4j2 and logback. The tradeoff between abstraction to remove code redundancy and flexibility in embedding the logic in the framework-specific plugin system is not big enough here IMO.

dreamorosi (approved):
Thanks for the super quick turnaround on this 🚀

phipag:
Thanks folks for your feedback. Let's bring this feature live in the next release 🚀 💚

Issue #2153, opened 2025-09-27
Feature request: Reuse namespace in Metrics.flushSingleMetric and other limitations


humanzz:
### Use case

As a user of Powertools v1, and considering moving to v2, I found the `withMetricsLogger` now replaced with `flushSingleMetric`. I see 2 issues
1. The current overloads, requires setting `namespace`, even though previously I've been relying on the namespace being whatever was already set (e.g. via powertools environment variables). I find this a degradation in UX if - which I think is a typical use case - I wanna use the same namespace
2. I also had use cases for `withMetricsLogger` where I add multiple metrics, which share an overriden dimensionset. Now this needs to be replaced with multiple calls to `flushSingleMetric`

I think the refactor/abstractions led to some degradations when it comes to `withMetricsLogger`. Moreover, `LambdaMetricsAspect` do set a few other properties/metadata that no longer seem to be available in instances emitted by `flushSingleMetric` such as trace and function id.

### Solution/User Experience

For 1, one solution would be to add `flushSingleMetric` overloads that don't take a `namespace`, and to update `EmfMetricsLogger` to treat the `namespace` parameter as nullable and a potential override to its `this.namespace`

For 2, maybe `Metrics` should get a new method `flushMetrics` with takes a `Consumer<Metrics>`, then one is more free about using default dimensions/namespace, or overriding them and adding multiple metrics.

### Alternative solutions

```Markdown
N/A
```

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda/java/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

leandrodamascena:
Hey @humanzz thanks for opening this issue and explaining more about the Developer Experience you expect from Powertools.

> ### Use case
> As a user of Powertools v1, and considering moving to v2, I found the `withMetricsLogger` now replaced with `flushSingleMetric`. I see 2 issues
> 
[...]

PR #2154, opened 2025-09-29
feat(metrics): introduce Metrics.flushMetrics

humanzz:
### Changes

> Please provide a summary of what's being changed

> Please add the issue number below, if no issue is present the PR might get blocked and not be reviewed

- introduce `Metrics.flushMetrics` as a more powerful version of `flushSingleMetrics` to allow
  - using defaults by inheriting state e.g. namespace, dimensions and properties
  - emitting multiple metrics in one metrics context
- update `EmfMetricsLogger.addMetadata` to use emf's `putProperty`
- refactor `flushSingleMetrics` and `captureColdStartMetric` to use `flushMetrics`
- consolidate request id and trace id setting into a newly introduced internal `MetricsUtils`

**Issue number:** #2153 

---

### DIFF
--- a/docs/core/metrics.md
+++ b/docs/core/metrics.md
@@ -484,8 +484,10 @@ You can create metrics with different configurations e.g. different namespace an
                 customMetrics.addMetric("CustomMetric", 1, MetricUnit.COUNT);
                 // To optionally set a different namespace
                 customMetrics.setNamespace("CustomNamespace");
-                // To optionally set different dimensions
-                customMetrics.setDefaultDimensions(DimensionSet.of("CustomDimension", "value"));
+                // To optionally set different default dimensions
+                customMetrics.setDefaultDimensions(DimensionSet.of("CustomDefaultDimension", "value"));
+                // To optionally append additional dimensions to default dimensions
+                customMetrics.addDimension(DimensionSet.of("CustomDimension", "value"));
                 // To optionally add metadata
                 customMetrics.addMetadata("CustomMetadata", "value"));
             });


powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
```diff
@@ -46,18 +43,6 @@ public final class MetricsFactory implements Resource {
...
-            String envNamespace = System.getenv("POWERTOOLS_METRICS_NAMESPACE");
-            if (envNamespace != null) {
-                metrics.setNamespace(envNamespace);
-            }
-
-            // Only set Service dimension if it's not the default undefined value
-            String serviceName = LambdaHandlerProcessor.serviceName();
-            if (!LambdaConstants.SERVICE_UNDEFINED.equals(serviceName)) {
-                metrics.setDefaultDimensions(DimensionSet.of("Service", serviceName));
-            }
```
phipag, line 60: I see in your PR description you moved this logic to `EmfMetricsLogger` to reduce the number of places that can set the namespace. I don't see how this reduces the number of places. In fact, it will increase the number of places if we add another metrics backend in addition to the EMF metrics logger.

Let's keep this in `MetricsFactory`. The idea of this architecture is that `EmfMetricsLogger` is a pluggable metrics backend that is only to be instantiated with an initial configuration by `MetricsFactory` or `MetricsBuilder`. This is also why it is in the `internal` package.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
```diff
@@ -50,15 +53,30 @@ public class EmfMetricsLogger implements Metrics {
...
     private String namespace;
     private Map<String, String> defaultDimensions = new HashMap<>();
+    private final Map<String, Object> metadata = new HashMap<>();
     private final AtomicBoolean hasMetrics = new AtomicBoolean(false);
 
     public EmfMetricsLogger(EnvironmentProvider environmentProvider, MetricsContext metricsContext) {
         this.emfLogger = new software.amazon.cloudwatchlogs.emf.logger.MetricsLogger(environmentProvider,
                 metricsContext);
         this.environmentProvider = environmentProvider;
+        this.metricsContext = metricsContext;
```
phipag: As the spotbugs finding already indicated, let's not keep a direct reference to this object here. I think we should not access it at all and this leads to a bug here ([see my comment below](https://github.com/aws-powertools/powertools-lambda-java/pull/2154/files#r2390228250)).

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
```diff
@@ -221,43 +240,22 @@ public void captureColdStartMetric(software.amazon.lambda.powertools.metrics.mod
...
-        // Create a new logger for this single metric
-        software.amazon.cloudwatchlogs.emf.logger.MetricsLogger singleMetricLogger = new software.amazon.cloudwatchlogs.emf.logger.MetricsLogger(
-                environmentProvider);
-
-        try {
-            singleMetricLogger.setNamespace(namespace);
-        } catch (Exception e) {
-            LOGGER.error("Namespace cannot be set for single metric due to an error in EMF", e);
+        // Create a new instance, inheriting namespace/dimensions state
+        EmfMetricsLogger metrics = new EmfMetricsLogger(environmentProvider, metricsContext);
```
phipag: Using the existing `metricsContext` will lead to a bug I believe. The `metricsContext` also keeps all the metrics. See implementation: 

```java
    public MetricsLogger putMetric(
            String key, double value, Unit unit, StorageResolution storageResolution)
            throws InvalidMetricException {
        rwl.readLock().lock();
        try {
            this.context.putMetric(key, value, unit, storageResolution);
            return this;
        } finally {
            rwl.readLock().unlock();
        }
    }
```

I believe this causes `flushMetrics` to not only flush the metrics added in the `Consumer` but also the metrics from the metrics logger whose configuration was inherited. Can you adjust your unit test to catch this behavior by adding a metric to the parent logger and asserting that this does not get flushed?

Let's create this new MetricsLogger from a fresh `new MetricsContext()` since we re-apply the configuration below anyway (still to be decided if this the way to go).

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
```diff
@@ -221,43 +240,22 @@ public void captureColdStartMetric(software.amazon.lambda.powertools.metrics.mod
...
+        if (namespace != null) {
+            metrics.setNamespace(this.namespace);
         }
+        defaultDimensions.forEach(metrics::addDimension);
+        metadata.forEach(metrics::addMetadata);
```
phipag: We haven't decided yet in the issue if we want to auto-inherit the default metrics logger configuration. I see in v1 this is the behavior we have.

spotbugs-exclude.xml
```diff
@@ -90,6 +90,10 @@
             <And>
                 <Class name="software.amazon.lambda.powertools.logging.internal.JsonSerializer"/>
             </And>
+            <And>
+                <Class name="software.amazon.lambda.powertools.metrics.internal.EmfMetricsLogger"/>
+                <Field name="metricsContext"/>
+            </And>
```
phipag: See my comment above. We can remove this exclusion again after implementing it.

phipag (changes requested):
Thanks @humanzz for sending this PR. Overall, it looks good and I left some comments that we need to address before moving forward.

Primarily, can you go back to the issue #2153 and let us know your use-case? We still need to decide whether we want to move forward with inheriting the default configuration of the metrics logger for `flushMetrics` and `flushSingleMetric`.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
humanzz: I did get the idea behind `MetricsFactory` and pluggable backends and I had some initial impressions that it might benefit from a bit more tightening. Now I had a look again at the factory and the builder, I think you're more right but let me share a couple of thoughts
- Initially, I thought that `EmfMetricsLogger` should use either the `MetricsProvider` or `MetricsFactory` to initialize a new instance of `Metrics` so it gets whichever default configurations (in this case environment variable namespace logic and the service name) which then contributed to moving more of the logic into the `EmfMetricsLogger`
- The other motivation, is that I see properties getting set on Metrics, spread across `LambdaMetricsAspect` and `EmfMetricsLogger` and `MetricsFactory`
- `MetricsFactory` is a bit odd in that it allows setting/overriding the `MetricsProvider` which can happen at any point in the lifetime of the class which can lead to issues if any piece of code decides to override the provider and other pieces rely on `getMetricsInstance` assuming it's all the global metrics instance that has been configured
- The fact that one now creates a `Metrics` instance in the handler (https://docs.powertools.aws.dev/lambda/java/latest/core/metrics/#creating-metrics) via the `MetricsFactory` or `MetricsBuilder` which would still under the hood setup/use the `MetricsFactory` where other non-handler code, if they want to use that instance would rely on `MetricsFactory.getMetricsInstance()` and hope/assume it's the same one defined in the lambda handler - while `setMetricsProvider` leaves the door open to overriding the provider is a bit problematic

I know some of the points above are not necessarily directly related to this PR but the thoughts came as I was trying to work on this.

I will return this piece to the `MetricsFactory`, and maybe for later, we can think about the points raised above

phipag: Thanks for this feedback about the implementation design. Let me explain some thoughts that went into it:
- The idea for `MetricsFactory` is to return a well configured (opinionated) `Metrics` backend Singleton. This should work independent of `LambdaMetricsAspect` i.e. independent of forcing the user to use AspectJ
- For users not using AspectJ, the `MetricsBuilder` can be used as an easy way to set the configuration on a default initialized `Metrics` backend returned by `MetricsFactory`
- For users using AspectJ with the `@FlushMetrics` annotation, the `LambdaMetricsAspect` only applies the configuration that is specific to the Aspect. 

All of this is designed in a way that is agnostic to the metrics backend and to achieve a consistent configuration precedence (https://docs.powertools.aws.dev/lambda/java/latest/core/metrics/#order-of-precedence-of-metrics-configuration). So, it is actually intentional that these different components all set configuration on the `Metrics` backend because the user is given different ways of configuring the metrics backend (Env vars, builder pattern, aspectj).

The idea of having the `setMetricsProvider` method was to allow users to:
1. Set another metrics backend programatically
2. Bring their own metrics provider implementation

Both of these scenarios are not documented yet.

I agree with your point that exposing this as a public method is not 100% ideal. Do you have a suggestion for a better way of achieving this?

humanzz:
I've addressed the `metricsContext` comments in https://github.com/aws-powertools/powertools-lambda-java/pull/2154/commits/4730be43a224a86621b69be3d6ce33f9c47a4190, and will leave details about the use case on the feature request.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
```diff
@@ -221,43 +224,22 @@ public void captureColdStartMetric(software.amazon.lambda.powertools.metrics.mod
...
-        try {
-            singleMetricLogger.setNamespace(namespace);
-        } catch (Exception e) {
-            LOGGER.error("Namespace cannot be set for single metric due to an error in EMF", e);
+        // Create a new instance, inheriting namespace/dimensions state
+        EmfMetricsLogger metrics = new EmfMetricsLogger(environmentProvider, new MetricsContext());
+        if (namespace != null) {
+            metrics.setNamespace(this.namespace);
         }
+        defaultDimensions.forEach(metrics::addDimension);
```
phipag: Should we use the `setDefaultDimensions` method here? 


```java
        if (!defaultDimensions.isEmpty()) {
            metrics.setDefaultDimensions(software.amazon.lambda.powertools.metrics.model.DimensionSet.of(defaultDimensions));
        }
```

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
humanzz: A couple of potential ideas
1. it's kept public, but it can be only set once, and trying to override it throws an exception
2. Using an approach like SPI to select the provider and completely dropping the ability to configure the provider via code - either for `MetricsFactory` or `MetricsBuilder`
3. Only allow `MetricsBuilder` to configure the provider, and it sets it on `MetricsFactory` using `setMetricsProvider` but make it package-private... but this still suffers from the issue that if someone initializes another `Metrics` using `MetricsBuilder` they might override the provider on the factory again - but we can prevent the overrides in a way similar to 1

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
phipag: Unrelated to this PR, I noticed this should not be updated for regular dimensions. Can you remove it? This is a bug.

powertools-metrics/src/test/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLoggerTest.java
```diff
@@ -446,6 +446,35 @@ void shouldReuseNamespaceForColdStartMetric() throws Exception {
                 .isEqualTo(customNamespace);
     }
 
+    @Test
+    void shouldFlushMetrics() throws Exception {
+        // Given
+        metrics.setNamespace("MainNamespace");
+        metrics.setDefaultDimensions(DimensionSet.of("CustomDim", "CustomValue"));
```
phipag, line 453: Can we add here:


```java
metrics.addDimension(DimensionSet.of("CustomDim2", "CustomValue2"));
```


and below assert

```java
assertThat(rootNode.get("CustomDim2")).isNull();
```

This tests for the bug where regular dimensions were accidentally discovered as default dimensions ([see comment](https://github.com/aws-powertools/powertools-lambda-java/pull/2154/files#r2390781166)).

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
phipag: I like these ideas. We use SPIs for e.g. idempotency or logging. In the current monolithic module setup for metrics I am not a fan of SPI though. Usually, we provide modules that auto-register. For example: `powertools-logging` will be used with `powertools-logging-log4j2` or `powertools-logging-logback`. 

Introducing SPIs will be a major version update and I will consider this when adding additional larger metrics backend. Great idea!

I really like option 1. It is simple enough and makes a lot of sense for such singleton classes – we don't want people to accidentally propagate a different metrics provider throughout the code base outside of initial configuration.

Would you be up for adding this in this PR including a small unit test?

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
humanzz: Sure... but now that am looking at this, why are dimension errors suppressed?
I think I'd rather get an exception over assuming metrics are emitted with correct dimensions but they're not?

humanzz:
@phipag another observation/inconsistency is the usage of `putProperty` vs. `putMetadata`. See
- https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java#L212
- https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspect.java#L95

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
phipag: The validation logic for the dimensions is already included in `DimensionSet`. It is not possible to submit invalid dimensions. 

We catch the underlying Emf logger exceptions to avoid having users rely on the exceptions raised by the logger – this is to leave the opportunity of dropping the emf library in favor of our own implementation.

phipag:
> @phipag another observation/inconsistency is the usage of `putProperty` vs. `putMetadata`. See
> 
>     * https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java#L212
> 
>     * https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspect.java#L95

Good catch! It seems that a "property" is simply a key added to the log line while "metadata" becomes part of the `_aws` object. These are logs from the [official example](https://github.com/humanzz/aws-lambda-powertools-java/blob/4730be43a224a86621b69be3d6ce33f9c47a4190/examples/powertools-examples-core-utilities/sam/src/main/java/helloworld/App.java#L58-L98):

**Cold start metric (putProperty)**

```json
{
    "_aws": {
        "Timestamp": 1751977621635,
        "CloudWatchMetrics": [
            {
                "Namespace": "ServerlessAirline",
                "Metrics": [
                    {
                        "Name": "ColdStart",
                        "Unit": "Count"
                    }
                ],
                "Dimensions": [
                    [
                        "Service",
                        "FunctionName"
                    ]
                ]
            }
        ]
    },
    "function_request_id": "b8628984-b745-44bb-8025-247c2da76da1",
    "xray_trace_id": "1-686d0e92-396456d519ba8c28174120a4",
    "ColdStart": 1,
    "Service": "payment",
}
```

**Regular metric (putMetadata)**

```json
{
    "_aws": {
        "Timestamp": 1751977946689,
        "CloudWatchMetrics": [
            {
                "Namespace": "ServerlessAirline",
                "Metrics": [
                    {
                        "Name": "CustomMetric1",
                        "Unit": "Count"
                    },
                    {
                        "Name": "CustomMetric3",
                        "Unit": "Count",
                        "StorageResolution": 1
                    }
                ],
                "Dimensions": [
                    [
                        "Service"
                    ]
                ]
            }
        ],
        "function_request_id": "feccb848-47a9-4b32-a16b-73d45d7ad308",
        "xray_trace_id": "1-686d0fdb-651f4fc05b57221e726890c0"
    },
    "CustomMetric1": 1,
    "Service": "payment",
    "CustomMetric3": 1
}
```

I need to find out the semantic difference – those nuances don't exist in the other runtimes. None of it is searchable in cloudwatch metrics. Besides the location in the json output I don't see any difference.


--- 

Update:

It appears that putMetadata will add key-value pairs to the `_aws` (metadata) object. Adding things there will not have an effect. Only `Timestamp` which is the officially documented metadata key in addition to `CloudWatchMetrics` (https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html#CloudWatch_Embedded_Metric_Format_Specification_structure_metadata%22%3ECloudWatch%20%20%20%20%20*%20%20%20%20%20Metadata). We shouldn't use this since it is nowhere documented and favor `putProperty`: https://github.com/awslabs/aws-embedded-metrics-java

We should update the `addMetadata` method to call `putProperty` on the EMF logger since this is the officially documented method (putMetadata is not documented). This is consistent with e.g. Python (see https://docs.powertools.aws.dev/lambda/python/latest/core/metrics/#add_metadata_outputjson)

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
```diff
@@ -92,6 +92,7 @@ public void addDimension(software.amazon.lambda.powertools.metrics.model.Dimensi
     @Override
     public void addMetadata(String key, Object value) {
         emfLogger.putMetadata(key, value);
```
phipag: Following [this comment](https://github.com/aws-powertools/powertools-lambda-java/pull/2154#issuecomment-3352176405) we should not call `putMetadata` but `putProperty`. It is the documented way and consistent with the other language runtimes. This will make the usage consistent again.

Thanks for catching this detail!

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
phipag: Do you think we should self-reference the `EmfMetricsLogger` class here instead of calling the underlying one directly to be consistent with your approach in `flushMetrics`?

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
humanzz: Yes. Was about the write this but you bet me to it :)

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
humanzz: Likely yes... so all is consistent

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
humanzz: One thought though, `Metrics.addMetada` and the new `Metrics.addProperty` which I'll introduce, are likely EMF-only features so there's some bleeding of the backend metrics engine to the abstraction.

phipag: I think we misunderstood each other. My suggestion was not add `addProperty` but instead replace the code within `addMetadata` with a call to `putProperty`. In the other Powertools runtimes we treat "metadata" as what is called in the Java EMF library "putProperty". 

Users should not call addMetadata on the EMF library at all. 

I also agree with your comment about bleeding. It is not possible to workaround this anymore without introducing a breaking change – in other backends this method will likely not have any effect.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
phipag: lgtm.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
humanzz: I missed the part about using `putProperty` in `addMetadata`

change coming!

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
```diff
@@ -221,43 +187,24 @@ public void captureColdStartMetric(software.amazon.lambda.powertools.metrics.mod
...
-
-        // Create a new logger for this single metric
-        software.amazon.cloudwatchlogs.emf.logger.MetricsLogger singleMetricLogger = new software.amazon.cloudwatchlogs.emf.logger.MetricsLogger(
-                environmentProvider);
-
-        try {
-            singleMetricLogger.setNamespace(namespace);
-        } catch (Exception e) {
-            LOGGER.error("Namespace cannot be set for single metric due to an error in EMF", e);
+        // Create a new instance, inheriting namespace/dimensions state
```
phipag:
```suggestion
        // Create a new instance, inheriting namespace, default dimensions, and metadata
```

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/LambdaMetricsAspect.java
```diff
@@ -90,11 +88,10 @@ public Object around(ProceedingJoinPoint pjp,
 
             metricsInstance.setRaiseOnEmptyMetrics(metrics.raiseOnEmptyMetrics());
 
-            // Add trace ID metadata if available
-            LambdaHandlerProcessor.getXrayTraceId()
-                    .ifPresent(traceId -> metricsInstance.addMetadata(TRACE_ID_PROPERTY, traceId));
+            Context extractedContext = extractContext(pjp);
+            MetricsUtils.addRequestIdAndXrayTraceIdIfAvailable(extractedContext, metricsInstance);
```
phipag, line 92: Good catch, the request id was missing here.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/MetricsUtils.java
```diff
@@ -0,0 +1,18 @@
+package software.amazon.lambda.powertools.metrics.internal;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import software.amazon.lambda.powertools.metrics.Metrics;
+
+import static software.amazon.lambda.powertools.common.internal.LambdaHandlerProcessor.getXrayTraceId;
+
+class MetricsUtils {
```
phipag: Can you make this class `final` and add a private constructor such as:

```java
private MetricsUtils {
    // Utility class
}
```
?

phipag: See Sonarlint finding and pmd analysis. 

```
For classes that only have static methods, consider making them utility classes. Note that this doesn't apply to abstract classes, since their subclasses may well include non-static methods. Also, if you want this class to be a utility class, remember to add a private constructor to prevent instantiation. (Note, that this use was known before PMD 5.1.0 as UseSingleton). UseUtilityClass (Priority: 1, Ruleset: Design) https://docs.pmd-code.org/snapshot/pmd_rules_java_design.html#useutilityclass
```

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/Metrics.java
```diff
@@ -162,7 +163,15 @@ default void captureColdStartMetric() {
     }
 
     /**
-     * Flush a single metric with custom dimensions. This creates a separate metrics context
+     * Flush a separate metrics context that inherits the namespace, dimensions and metadata. This creates a separate metrics context
+     * that doesn't affect the default metrics context.
```
phipag:
```suggestion
     * Flush a separate metrics context that inherits the namespace, default dimensions, and metadata. This creates a separate metrics context
     * that doesn't affect the default metrics context.
```

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/Metrics.java
```diff
@@ -162,7 +163,15 @@ default void captureColdStartMetric() {
...
+     *
+     * @param metricsConsumer the consumer to use to edit the metrics instance (e.g. add metrics, override namespace) before flushing
```
phipag:
```suggestion
     * @param metricsConsumer the consumer to use to edit the metrics instance (e.g. add metrics, override namespace, add custom dimensions) before flushing
```

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/Metrics.java
```diff
@@ -183,6 +199,9 @@ default void captureColdStartMetric() {
      * @param namespace  the namespace for the metric
      */
     default void flushSingleMetric(String name, double value, MetricUnit unit, String namespace) {
-        flushSingleMetric(name, value, unit, namespace, null);
+        flushMetrics(metrics -> {
+            metrics.setNamespace(namespace);
+            metrics.addMetric(name, value, unit);
+        });
     }
```
phipag, line 206: Following the discussion in the issue: All runtimes inherit the namespace by default if it is not overwritten. Let's add the method overload you suggested in your initial feature request here (without namespace param). We should also add a unit test for this.

```java
    default void flushSingleMetric(String name, double value, MetricUnit unit) {
        flushMetrics(metrics -> {
            metrics.addMetric(name, value, unit);
        });
```

phipag (commented):
Thanks @humanzz for your continued engagement. Everything looks good so far besides these minor comments. 🚀 

The final step after this will be to update the documentation with the new `flushMetrics` method including code examples showing developers how to use it. Let me know if you would like to do this – otherwise I am happy to commit the documentation into this PR. 

The docs file uses [mkdocs](https://www.mkdocs.org/) and is located here `docs/core/metrics.md`.

powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/Metrics.java
humanzz: While this might make sense, a couple of thoughts
1. Not sure what value this would provide over `Metrics.addMetric` (I understand it's a different flush, but end result is that the same metric is emitted, reusing namespace, dimensions, metadata)
2. I sincerely believe that all `flushSingleMetric` should be marked as deprecated, to help reduce the api surface area as `flushMetrics` would already enable all that's supported by these overloads

phipag: Interesting thought. Let's not add the additional overload for now. I'll think about deprecating these methods in a future v3 release. Let's not decide this in the scope of this PR.

docs/core/metrics.md
```diff
@@ -480,13 +480,15 @@ You can create a single metric with its own namespace and dimensions using `flus
...
-                MetricUnit.COUNT,
-                "CustomNamespace",
-                DimensionSet.of("CustomDimension", "value")  // Dimensions are optional
-            );
+            metrics.flushMetrics((customMetrics) -> {
+                customMetrics.addMetric("CustomMetric", 1, MetricUnit.COUNT);
+                // To optionally set a different namespace
+                customMetrics.setNamespace("CustomNamespace");
+                // To optionally set different dimensions
+                customMetrics.setDefaultDimensions(DimensionSet.of("CustomDimension", "value"));
```
phipag:
```suggestion
                // To optionally set different default dimensions
                customMetrics.setDefaultDimensions(DimensionSet.of("CustomDefaultDimension", "value"));
                // To optionally append additional dimensions to default dimensions
                customMetrics.addDimension(DimensionSet.of("CustomDimension", "value"));
```

phipag (commented):
Awesome! Final small comment (you can commit directly from GitHub UI I think).

phipag (approved):
Thank you @humanzz for raising the bar. Your attention to detail significantly improved this module! 

These changes look good to me. Let me align with the other runtimes and make a merge decision! 🚀

phipag:
Aligned. I am merging this PR. Congrats @humanzz on this awesome contribution! 🎉

Issue #2204, opened 2025-10-17
Feature request: Support functional interface for Logging utility


phipag:
### Use case

The Logging utility currently only offers a benefit to developers when used with the AspectJ `@Logging` annotation. The purpose of this feature request is to add a functional interface to initialize the Logging utility similar to `TracingUtils`. 

### Solution/User Experience

We can re-use the existing `PowertoolsLogging` facade that abstracts away the logging backend for certain Powertools specific features such as Log buffering.

**Example**

Usage with AspectJ
```java
public class App implements RequestHandler<APIGatewayProxyRequestEvent, String> {
    private static final Logger log = LoggerFactory.getLogger(App.class);

    @Logging(logEvent = true, logResponse = true, samplingRate = 0.7, correlationIdPath = "requestContext.requestId")
    public String handleRequest(APIGatewayProxyRequestEvent input, Context context) {
        log.info("Processing request");
        return "Hello World";
    }
}

```

Usage without AspectJ
```java
public class App implements RequestHandler<APIGatewayProxyRequestEvent, String> {
    private static final Logger log = LoggerFactory.getLogger(App.class);

    public String handleRequest(APIGatewayProxyRequestEvent input, Context context) {
        PowertoolsLogging.initializeLogging(context, 0.7, "requestContext.requestId", input);
        
        // Manual event logging (if desired)
        log.info("Handler Event", entry("event", input));
        
        log.info("Processing request");
        String response = "Hello World";
        
        // Manual response logging (if desired)
        log.info("Handler Response", entry("response", response));
        
        return response;
    }
}
```

### Alternative solutions

```Markdown

```

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda/java/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

phipag:
Created a first draft in https://github.com/aws-powertools/powertools-lambda-java/pull/2205

Issue #2202, opened 2025-10-17
Feature request: Middleware Chain API for Powertools Utilities


phipag:
## Use case

Currently, Powertools for AWS Lambda (Java) provides individual utilities like Logging, Tracing, Metrics, and other features through AspectJ annotations and static methods. While effective, this approach lacks a unified and modern functional interface for applying multiple Powertools utilities to Lambda handlers without relying on AspectJ annotations.

Developers need a solution that offers:
- A unified functional API to compose multiple Powertools features (logging, tracing, metrics, validation, etc.)
- Type-safe middleware composition that works across all Lambda handler types (RequestHandler, RequestStreamHandler, etc.)
- Clean separation between business logic and cross-cutting concerns without AspectJ dependencies
- Flexible composition allowing different orders and combinations of middlewares
- Modern Java functional programming patterns that feel idiomatic to Java developers

This would provide an alternative to the current AspectJ annotation-based approach while maintaining the same powerful utility features.

## Solution/User Experience

Introduce a middleware chain API that allows functional composition of Powertools features. The design below suggests an implementation of the Chain of Responsibility software design pattern:

```java
@FunctionalInterface
public interface Middleware<T, R> {
    R apply(T input, Context context, BiFunction<T, Context, R> next);
}

public class MiddlewareChain<T, R> {
    public MiddlewareChain<T, R> use(Middleware<T, R> middleware) { /* ... */ }
    public R execute(T input, Context context, BiFunction<T, Context, R> handler) { /* ... */ }
}

public class PowertoolsMiddlewares {
    public static <T, R> Middleware<T, R> logging() { /* ... */ }
    public static <T, R> Middleware<T, R> tracing(String serviceName) { /* ... */ }
    public static <T, R> Middleware<T, R> metrics(String namespace) { /* ... */ }
}
```

Usage example:
```java
public class OrderHandler implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
    
    private final MiddlewareChain<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> chain = 
        new MiddlewareChain<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent>()
            .use(PowertoolsMiddlewares.logging())
            .use(PowertoolsMiddlewares.tracing("OrderService"))
            .use(PowertoolsMiddlewares.metrics("Orders"));
    
    @Override
[...]

PR #2205, opened 2025-10-17
feat(logging): Support functional interface in addition to AspectJ `@Logging` annotation

phipag:
## Summary

This PR introduces an alternative functional interface to the Logging utility that does not depend on the AspectJ `@Logging` annotation. We re-use `PowertoolsLogging` as entry point for the Powertools logging experience which exposes two new methods:
1. `initializeLogging()` – has some method overloads for convenience
2. `clearState()` – can optionally clear the logging context as well

This is also thread-safe now. 

We also add a new E2E test asserting that it yields the exact same outputs as the AspectJ aspect based way.

The before and after compared to AspectJ looks like this:

Usage with AspectJ
```java
public class App implements RequestHandler<APIGatewayProxyRequestEvent, String> {
    private static final Logger log = LoggerFactory.getLogger(App.class);

    @Logging(logEvent = true, logResponse = true, samplingRate = 0.7, correlationIdPath = "requestContext.requestId")
    public String handleRequest(APIGatewayProxyRequestEvent input, Context context) {
        log.info("Processing request");
        return "Hello World";
    }
}

```

Usage without AspectJ
```java
public class App implements RequestHandler<APIGatewayProxyRequestEvent, String> {
    private static final Logger log = LoggerFactory.getLogger(App.class);

    public String handleRequest(APIGatewayProxyRequestEvent input, Context context) {
        try {
                PowertoolsLogging.initializeLogging(context, 0.7, "requestContext.requestId", input);
                
                // Manual event logging (if desired)
                log.info("Handler Event", entry("event", input));
                
                log.info("Processing request");
                String response = "Hello World";
                
                // Manual response logging (if desired)
                log.info("Handler Response", entry("response", response));
                
                return response;
        } finally {
            PowertoolsLogging.clearState()
        }
    }
}
```

Note: In this small RFC I am investigating an improvement to this imperative API leveraging a more modern functional Middleware approach similar to TypeScript middy.js. See https://github.com/aws-powertools/powertools-lambda-java/issues/2202

### Changes

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/2204

---

### DIFF
--- a/powertools-e2e-tests/handlers/logging-functional/src/main/java/software/amazon/lambda/powertools/e2e/Function.java
+++ b/powertools-e2e-tests/handlers/logging-functional/src/main/java/software/amazon/lambda/powertools/e2e/Function.java
@@ -21,21 +21,20 @@ import org.slf4j.MDC;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 
-import software.amazon.lambda.powertools.logging.Logging;
 import software.amazon.lambda.powertools.logging.PowertoolsLogging;
 
 public class Function implements RequestHandler<Input, String> {
     private static final Logger LOG = LoggerFactory.getLogger(Function.class);
 
     public String handleRequest(Input input, Context context) {
-        PowertoolsLogging.initializeLogging(context);
+        return PowertoolsLogging.withLogging(context, () -> {
+            input.getKeys().forEach(MDC::put);
+            LOG.info(input.getMessage());
 
-        input.getKeys().forEach(MDC::put);
-        LOG.info(input.getMessage());
+            // Flush buffer manually since we buffer at INFO level to test log buffering
+            PowertoolsLogging.flushBuffer();
 
-        // Flush buffer manually since we buffer at INFO level to test log buffering
-        PowertoolsLogging.flushBuffer();
-
-        return "OK";
+            return "OK";
+        });
     }
 }
--- a/powertools-logging/src/main/java/software/amazon/lambda/powertools/logging/PowertoolsLogging.java
+++ b/powertools-logging/src/main/java/software/amazon/lambda/powertools/logging/PowertoolsLogging.java
@@ -30,6 +30,7 @@ import java.util.Arrays;
 import java.util.Locale;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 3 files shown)


hjgraca:
Just one small observation on the API design and developer experience. 
Unfortunately we need to have `try... finally ` block, but will developers consistently remember to call clearState() in finally blocks? 

Could we add to the API the possibility to create a functional wrapper that handles the try-finally automatically? I think this is what you mention in the #2202 issue

For example:

```java
PowertoolsLogging.withLogging(context, config, () -> {
    log.info("Processing request");
    return "Hello World";
});
```

powertools-logging/pom.xml
```diff
@@ -186,6 +186,7 @@
                 <configuration>
                     <environmentVariables>
                         <AWS_LAMBDA_LOG_FORMAT>JSON</AWS_LAMBDA_LOG_FORMAT>
+                        <POWERTOOLS_SERVICE_NAME>testService</POWERTOOLS_SERVICE_NAME>
```
hjgraca, line 189: Is this change for the tests?

phipag: Yes, it was missing in `powertools-logging` but included in `powertools-logging-log4j2` and `powertools-logging-logback`. This caused flaky tests if the order of execution of tests differed. If `powertools-logging` tests were executed **before** the sub-modules tests passed. If it got executed **after** the test modules the environment variables was set leading to a failing assertion where we asserted against `service_undefined`. Therefore, I made it consistent with the sub-modules to make the tests pass consistently.

phipag:
> Just one small observation on the API design and developer experience. Unfortunately we need to have `try... finally ` block, but will developers consistently remember to call clearState() in finally blocks?
> 
> Could we add to the API the possibility to create a functional wrapper that handles the try-finally automatically? I think this is what you mention in the #2202 issue
> 
> For example:
> 
> ```java
> PowertoolsLogging.withLogging(context, config, () -> {
>     log.info("Processing request");
>     return "Hello World";
> });
> ```

Yes, I like this idea. I'll implement such a wrapper function. 

Correct, issue #2202 will abstract such wrappers away from individual utilities to a middleware approach which is more scalable. But I like the idea of having a functional wrapper for an individual utility as well for customers who simply want to get started with a utility.

phipag:
Thanks @hjgraca. I implemented your suggestions. See this file for how it is used in E2E tests: https://github.com/aws-powertools/powertools-lambda-java/pull/2205/files#diff-3bcb5e87cb7cc4655b7d9427e50b10a9546c2a92910f6ae860635a041db2ebc9

Issue #2077, opened 2025-08-22
Maintenance: Add workflow to update changelog in GitHub and documentation


phipag:
### Summary

Similar to other repositories such as https://github.com/aws-powertools/powertools-lambda-typescript we should include a GitHub workflow that automatically updates the CHANGELOG.md on GitHub and also the documentation on each version release.

### Why is this needed?

Currently, the changelog is not kept up-to-date impacting the customer experience!

### Which area does this relate to?

Automation

### Solution

We should investigate using this GH action: https://github.com/aws-powertools/actions/tree/main/.github/actions/version-n-changelog

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

aryannikhil:
@phipag I’m interested in contributing to this project. Could you please assign it to me?

phipag:
Hey @nikkhilaaryan, I think for this one it is better for a maintainer to pickup due to the permissions required for GitHub actions and a quicker testing feedback loop due to this fact. I assigned this issue to myself to reflect that.

I would like to encourage you to checkout any issues marked with `good-first-issue` issue label that is not yet assigned:  https://github.com/aws-powertools/powertools-lambda-java/issues?q=is%3Aissue%20state%3Aopen%20label%3Agood-first-issue

Let me know if you have any questions.

leandrodamascena:
Hey @sthulb  Pls work in this issue with @phipag

phipag:
PR merged. I will test it in the release today.

phipag:
There is an issue: https://github.com/aws-powertools/powertools-lambda-java/actions/runs/18680917425/job/53262600051. 

The docker sha256 might be incorrectly formatted.

phipag:
Merged a PR to fix #2218. Retrying release workflow now.

phipag:
git tag is missing for Changelog. https://github.com/aws-powertools/powertools-lambda-java/actions/runs/18682982408/job/53269178397

phipag:
New PR: https://github.com/aws-powertools/powertools-lambda-java/pull/2219

phipag:
[...]

PR #2207, opened 2025-10-17
chore(ci): fix changelog generation

sthulb:
### Changes

> Please provide a summary of what's being changed

> Please add the issue number below, if no issue is present the PR might get blocked and not be reviewed

**Issue number:** #2077

---

### DIFF
--- a/.github/workflows/security-scorecard.yml
+++ b/.github/workflows/security-scorecard.yml
@@ -52,6 +52,6 @@ jobs:
           path: results.sarif
           retention-days: 5
       - name: Upload to Code-Scanning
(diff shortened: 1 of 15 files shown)


.github/workflows/release.yml
```diff
@@ -240,9 +240,10 @@ jobs:
           git config pull.rebase true
           git config remote.origin.url >&-
       - id: branch
-        name: Create branch
+        name: Create branch and update change log
         run: |
           git checkout -b ci-${{ github.run_id }}
+          docker run -v "${PWD}":/workdir quay.io/git-chglog/git-chglog:0.15.1 > CHANGELOG.md
```
phipag: Do we need to use the image hash instead of the version number here?

sthulb: this was lifted directly from Python's workflow – but maybe

Issue #2221, opened 2025-10-21
Docs: Document usage of functional API for all utilities


phipag:
### What were you searching in the docs?

After adding functional APIs for all utilities as part of this parent issue (https://github.com/aws-powertools/powertools-lambda-java/issues/2202) we should update our documentation to reflect these changes.

### How can we improve?

This issue suggests to create a "Usage Patterns" page similar to the TypeScript documentation (https://docs.aws.amazon.com/powertools/typescript/latest/getting-started/usage-patterns/). 

Additionally, the individual pages of utilities should be update to highlight this new possibility of using a functional API. We can mention this during the installation instructions highlighting that AspectJ is only needed when using annotations but not for the functional interface.

### Acknowledgment

- [x] I understand the final update might be different from my proposed suggestion, or refused.

Issue #2213, opened 2025-10-21
Feature request: Support functional interface for Idempotency utility


phipag:
### Use case

The Idempotency utility currently only offers a benefit to developers when used with the AspectJ @Idempotent annotation and `@IdempotencyKey` annotation. The purpose of this feature request is to add a functional interface to apply the Idempotency utility independent of AspectJ to both the Lambda handler and also arbitrary methods called by the Lambda handler. For example, by wrapping it in a higher order function.

### Solution/User Experience

The proposed design is below. We need make sure that this is **thread-safe** since this can be combined with the parallel batch processor as well. Additionally, we need to be sure that reading the `@IdempotencyKey` reflectively does not impact performance – we should consider caching the key locally upon retrieving it with reflection.

Usage with AspectJ

```java
public class PaymentHandler implements RequestHandler<SQSEvent, List<String>> {

    public PaymentHandler() {
        Idempotency.config()
                .withPersistenceStore(
                        DynamoDBPersistenceStore.builder()
                                .withTableName(System.getenv("TABLE_NAME"))
                                .build())
                .configure();
    }

    @Override
    public List<String> handleRequest(SQSEvent sqsEvent, Context context) {
        Idempotency.registerLambdaContext(context);
        return sqsEvent.getRecords().stream()
            .map(record -> process(record.getMessageId(), record.getBody()))
            .collect(Collectors.toList());
    }

    @Idempotent
    private String process(String messageId, @IdempotencyKey String messageBody) {
        logger.info("Processing messageId: {}", messageId);
        PaymentRequest request = extractDataFrom(messageBody).as(PaymentRequest.class);
        return paymentService.process(request);
    }
}

```

Usage without AspectJ. `@IdempotencyKey` is read using reflection and cached to avoid performance problems when used e.g. in batch processors / parallel batch processor. The Cache needs to be thread-safe ideally with minimal locking.

```java
public class PaymentHandler implements RequestHandler<SQSEvent, List<String>> {

    public PaymentHandler() {
        Idempotency.config()
                .withPersistenceStore(
                        DynamoDBPersistenceStore.builder()
                                .withTableName(System.getenv("TABLE_NAME"))
                                .build())
                .configure();
    }

    @Override
    public List<String> handleRequest(SQSEvent sqsEvent, Context context) {
        Idempotency.registerLambdaContext(context);
        return sqsEvent.getRecords().stream()
            .map(record -> PowertoolsIdempotency.makeIdempotent(
[...]

PR #2244, opened 2025-10-28
feat(idempotency): Add functional `Idempotency` API

phipag:
## Summary

This PR introduces a new functional API for the Idempotency utility based on the design of the issue below. It was **not** possible to implement the `@IdempotencyKey` based design. If a method reference (or Lambda function) is passed to the `makeIdempotent` function it is not possible to find the annotation using reflection at runtime. Therefore, the idempotencyKey needs to be passed explicitly for multi-parameter functions. The same is true for the return type of the method – otherwise deserialization of an idempotent record fails. This is the trade-off of the functional API.

This PR also makes Idempotency thread-safe.

A **new feature** was added. Using the functional interface, we can now support making functions returning generic types idempotent. This was not possible using AspectJ.

Respective overloads were created to cover all use cases:

**1. Simple Function Reference (Single Parameter)**

```java
public Basket handleRequest(Product input, Context context) {
    Idempotency.registerLambdaContext(context);
    // Basket is the return type of the idempotent function ("processProduct")
    return Idempotency.makeIdempotent(this::processProduct, input, Basket.class);
}

private Basket processProduct(Product product) {
    // business logic
}
```

**2. Supplier with Idempotency Key (Multi-Parameter Methods)**

```java
public List<String> handleRequest(SQSEvent event, Context context) throws Throwable {
    Idempotency.registerLambdaContext(context);
    
    return event.getRecords().stream()
        .map(record -> {
            return Idempotency.makeIdempotent(
                record.getBody(),
                () -> processPayment(record.getMessageId(), record.getBody()),
                String.class
            );
        })
        .collect(Collectors.toList());
}

```

**3. Explicit Function Name (Same Payload, Different Operations)**

```java
public Response handleRequest(Order order, Context context) {
    Idempotency.registerLambdaContext(context);
    
    // Same orderId, different operations
    Idempotency.makeIdempotent("processPayment", order.getId(), 
        () -> processPayment(order), String.class);
    
    Idempotency.makeIdempotent("sendConfirmation", order.getId(), 
        () -> sendEmail(order), String.class);
    
    return new Response("success");
}
```

**4. Generic Return Types (TypeReference Support)**

```java
public Map<String, Basket> handleRequest(Product input, Context context) {
    Idempotency.registerLambdaContext(context);
    
    return Idempotency.makeIdempotent(
        this::processProduct,
        input,
        new TypeReference<Map<String, Basket>>() {}
    );
}

private Map<String, Basket> processProduct(Product product) {
    // business logic returning generic type
}
```

Especially use-case 3 needs to be documented well (separate issue here #2221). This is needed because now we cannot infer the `functionName` anymore that is used to de-duplicate the same payload for different function names.

### Changes

**For the reviewers of this PR:**

Ignore the PMD finding "A catch statement should never catch throwable since it includes errors.". It is not possible to fix without breaking change. I hide this from the new customer facing API already but it is needed to keep AspectJ compatible.

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/2213

---

### DIFF
--- a/powertools-idempotency/powertools-idempotency-core/src/main/java/software/amazon/lambda/powertools/idempotency/Idempotency.java
+++ b/powertools-idempotency/powertools-idempotency-core/src/main/java/software/amazon/lambda/powertools/idempotency/Idempotency.java
@@ -143,7 +143,7 @@ public final class Idempotency {
         this.persistenceStore = persistenceStore;
     }
 
-    private static class Holder {
+    private static final class Holder {
         private static final Idempotency instance = new Idempotency();
     }
 


powertools-e2e-tests/handlers/idempotency-functional/src/main/java/software/amazon/lambda/powertools/e2e/Function.java
```diff
@@ -0,0 +1,67 @@
...
+                .configure();
+    }
+
+    public String handleRequest(Input input, Context context) {
+        Idempotency.registerLambdaContext(context);
+
+        return PowertoolsIdempotency.makeIdempotent(this::processRequest, input, String.class);
+    }
+
+    private String processRequest(Input input) {
```
hjgraca, line 62: Avoid unused method parameters such as 'input'. error detected

phipag: This is actually a wrong finding. 

```java
return PowertoolsIdempotency.makeIdempotent(this::processRequest, input, String.class);
```

`makeIdempotent` will auto-call the `processRequest` passing `input` as argument.

hjgraca (dismissed):
GTM great work

Issue #2251, opened 2025-10-31
Bug: Jackson Joda time needed for some AWS events


phipag:
### Expected Behaviour

When logging handler events or attempting to extract the correlation ID from an event such as `ScheduledEvent`, I expect that the event gets serialized to a valid JSON when logging and that the correlation ID extraction is working as expected.

### Current Behaviour

For AWS events that use `Joda` time under the hood, for example [`ScheduledEvent`](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java), we see the following exception:

```
Joda date/time type `org.joda.time.DateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-joda" to enable handling (through reference chain: com.amazonaws.services.lambda.runtime.events.ScheduledEvent["time"]): java.lang.IllegalArgumentException
java.lang.IllegalArgumentException: Joda date/time type `org.joda.time.DateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-joda" to enable handling (through reference chain:
```

The correlation ID is not extracted at all. The event logging falls back to calling `toString()` on the event. In case of the `ScheduledEvent` this yields an invalid json due to a non-JSON compliant implementation in the AWS event library. 

Example:

```
{
    "level": "INFO",
    "message": "Handler Event",
    "cold_start": true,
    "thread": "main",
    "function_arn": "arn:aws:lambda:us-east-1:<ACCOUNT_ID>:function:powertools-examples-batch-sq-DemoSQSSenderFunction-kGvQ7guWRsW0",
    "function_memory_size": 5400,
    "function_name": "powertools-examples-batch-sq-DemoSQSSenderFunction-kGvQ7guWRsW0",
    "function_request_id": "d780b6fc-d30a-497c-a75c-400bea718983",
    "function_version": "$LATEST",
    "service": "sqs-batch-demo",
    "timestamp": "2025-10-31T09:57:33.526Z",
    "xray_trace_id": "1-6904880a-3b7f5d6d0dc62d10320adf78",
[...]

PR #2252, opened 2025-10-31
fix(serialization): Fix Joda time serialization for AWS events.

phipag:
## Summary

Some AWS Events still use Joda time. We need to add the Joda time Jackson module to be able to serialize them as valid json during `@Logging(logEvent = true)`. Additionally, it is needed to be able to extract correlation IDs from these events.

This PR adds the Joda time module to `powertools-serialization` which is used across the library (e.g. in Logging) for all serialization purposes.

### Changes

**Issue number:** #2251 

---

### DIFF
--- a/powertools-serialization/src/main/java/software/amazon/lambda/powertools/utilities/JsonConfig.java
+++ b/powertools-serialization/src/main/java/software/amazon/lambda/powertools/utilities/JsonConfig.java
@@ -25,6 +25,7 @@ import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.json.JsonMapper;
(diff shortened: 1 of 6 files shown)


powertools-serialization/src/main/java/software/amazon/lambda/powertools/utilities/JsonConfig.java
```diff
@@ -41,11 +42,13 @@ public final class JsonConfig {
             // Don't throw an exception when json has extra fields you are not serializing on.
             .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
             // Ignore null values when writing json.
-            .serializationInclusion(JsonInclude.Include.NON_NULL)
+            .defaultPropertyInclusion(
+                    JsonInclude.Value.construct(JsonInclude.Include.NON_NULL, JsonInclude.Include.USE_DEFAULTS))
```
phipag, line 46: Unrelated to this change but `.serizliationInclusion` method is deprecated.

Issue #2289, opened 2025-11-14
Maintenance: Thread-safety for Metrics Utility


phipag:
### Summary

The Metrics utility can be used from within the parallel batch processor. The current operating model is:
- Create `Metrics` singleton using `MetricsFactory` or `MetricsBuilder`
- Use it from anywhere within your Lambda handler
- Flush once when the handler finishes (or max batch size is reached)

The problem in multi-threaded scenarios is that metrics configurations such as dimensions / metadata or similar will leak across threads even when flushing manually from within a thread.

### Why is this needed?

This is needed to propose the expected thread-safety quality bar to customers.

### Which area does this relate to?

Metrics

### Solution

Use `InheritableThreadLocal` to manage Metrics instances. 

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

phipag:
Resolved by #2294

PR #2294, opened 2025-11-14
improv(metrics): Add thread-safety to Metrics utility through thread-local Metrics instance management

phipag:
## Summary

This PR makes the Metrics utility thread-safe for scenarios when used as a singleton in a Lambda handler that uses multi-threading while modifying dimensions or other metadata from within threads.

### Changes

The main change is the introduction of a `RequestScopedMetricsProxy`. This class lazily initializes the corresponding metrics backend on the first invocation of e.g. `.addMetric(...)` using a `ConcurrentHashMap<String, Metrics>` keyed by X-Ray trace ID. This ensures that concurrent usages of Metrics are isolated per Lambda invocation (not per thread), and metrics added in child threads are correctly flushed.

This is the same approach we use in log buffering. See for example:

https://github.com/aws-powertools/powertools-lambda-java/blob/4e83d8a3b5cfbdb26d6ae5db81393d9c47c78df5/powertools-logging/powertools-logging-log4j/src/main/java/software/amazon/lambda/powertools/logging/log4j/BufferingAppender.java#L102-L103

https://github.com/aws-powertools/powertools-lambda-java/blob/4e83d8a3b5cfbdb26d6ae5db81393d9c47c78df5/powertools-logging/powertools-logging-log4j/src/main/java/software/amazon/lambda/powertools/logging/log4j/BufferingAppender.java#L112-L124

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/2289

---

### DIFF
--- a/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/RequestScopedMetricsProxy.java
+++ b/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/RequestScopedMetricsProxy.java
@@ -18,6 +18,7 @@ import java.time.Instant;
 import java.util.HashMap;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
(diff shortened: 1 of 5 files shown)


powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java
```diff
@@ -44,23 +46,23 @@ public final class MetricsFactory implements Resource {
      * @return the singleton Metrics instance
      */
     public static synchronized Metrics getMetricsInstance() {
-        if (metrics == null) {
-            metrics = provider.getMetricsInstance();
+        if (metricsProxy == null) {
+            metricsProxy = new RequestScopedMetricsProxy(provider);
```
svozza, line 50: Nice...

svozza (approved):

Issue #2263, opened 2025-11-04
Maintenance: Build Java 25 in GitHub actions


phipag:
### Summary

We should start building and testing for Java 25 in GitHub actions for both JVM and native tests and builds. 

### Why is this needed?

This is needed to make sure that we support also the most recent Java features.

### Which area does this relate to?

Governance

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda-java/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/)

phipag:
Merged PR https://github.com/aws-powertools/powertools-lambda-java/pull/2285. 

This PR enables Java 25 builds for JVM. 

Next steps:
- Enable Java 25 builds for GraalVM native tests (either via 21, 25 matrix or drop support for GraalVM 21)
- Enable Java 25 E2E tests

dreamorosi:
Hi @phipag, I'm confused by the PR name and next steps in the message above.

The title of the PR says that we've enabled Java 25 E2E tests, but there's a next step with the same words.

What is there left to do and what has been done? Also, how big are these unit of work? Do they need their dedicated sub-issues?

phipag:
Hey @dreamorosi, 
I forgot to close this issue. The last PR already resolved all next steps. We could not enable the E2E tests last week due to the Java 25 runtime not being available yet which is why had to send 2 separate PRs. 

Closing this issue now!

PR #2298, opened 2025-11-17
chore(ci): Enable Java 25 E2E tests on Java 25 Lambda runtime.

phipag:
## Summary

This PR enables Java 25 for E2E tests deploying on the new managed Java 25 Lambda runtime. 

We also `RETAIN` log groups to allow for easier debugging of failed tests (log groups have a retention period of 1 day for E2E tests to make sure of cleanup).

We also run against GraalVM 25 only for the native E2E tests since it is not necessary to test different JVM versions when compiled to a native image with using Graal. This saves runner resources.

### Changes

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/2263

---


phipag:
The E2E tests work on Java 25 JVM but the GraalVM 25 tests were failing due to missing unsafeAllocate flags in the config. Retrying the tests with this new config option now.

Issue #1671, opened 2024-06-25
Propagate traces and log context to non-main threads


jeromevdl:
Powertools for AWS Lambda (Java) should support multithreading properly, especially but not-only:
- with log context and MDC, so that custom keys are not lost between threads
- with traces and X-Ray, so that subsegment have the good parent segment when changing thread

Note that these 2 components use `ThreadLocal`.

**What were you trying to accomplish?**

## Expected Behavior
- X-Ray traces should have the good hierarchy: subsegments within threads should be under segment within 'main' thread
- MDC keys (for logs) should be kept in all threads.

## Current Behavior
- Using the new feature (#1620), we can see in X-Ray that thread sub-segments don't have the good parent (handler).
- MDC keys are lost between threads

## Steps to Reproduce (for bugs)
1. Use the [powertools-examples-batch](https://github.com/aws-powertools/powertools-lambda-java/tree/main/examples/powertools-examples-batch) with SQS
2. See traces:
![image](https://github.com/aws-powertools/powertools-lambda-java/assets/117538/baa4b3df-ee2d-492a-86fe-605b293610b3)

## Possible Solution
For logs, a temporary solution has been implemented ([`MultiThreadMDC`](https://github.com/aws-powertools/powertools-lambda-java/pull/1620/files#diff-0d80cb0adc69c617395eef88678c4d1cc7885ad5939cbb861cf77004d35d443c).

We should think about a more "generic" or "universal" way of managing multithreading in Powertools to not only make logging and tracing work, but also enable users to use multithread more simply.

Initial thoughts:
- Having an `ThreadAware` interface in the common module and a static list of `ThreadAware` classes. 
- `ThreadAware` interface should have at least 2 methods to act before the thread starts and within the thread execution, to be able to copy data from `main` thread (or perform another task like getting the parent x-ray segment) and reuse it within the thread.
- When working with thread, we should iterate over this list and call these methods. Maybe a util class in common could simplify this too.

## Environment

* **Powertools for AWS Lambda (Java) version used**: 2.0.0-SNAPSHOT

scottgerring:
x-ray specific #1670

humanzz:
[...]

PR #2300, opened 2025-11-19
improv(batch): Propagate trace entity to worker threads during parallel processing.

phipag:
## Summary

This PR fixes an issue when using `@Tracing` together with parallel batch processing.

**Before**

When using parallel batch processing (without passing a custom `Executor`) we use a `parallelStream()` under the hood. A `parallelStream()` uses the common fork-join pool plus the main thread for processing. What happened was that some of sub-segments were not correctly attached to the parent `## handleRequest` sub-segment. The reason is that X-RAY SDK maintains thread local trace entity instances. Hence, only the sub-sgements from the main thread and not those of the fork-join pool threads were correctly attached to the parent segment leading to a tracing output like this:

Screenshot 2025-11-18 at 15 12 28

**After**

This PR introduces propagation of the main thread's trace entity to the worker threads of the parallel batch processing worker threads. This fixes the issue such that all sub-segments are correctly attached to their parent segments. See screenshot after:

image (8)

### Changes

Added `XRayTraceEntityPropagator` as a reflection wrapper around the X-RAY SDK methods needed for trace propagation following documentation at https://docs.aws.amazon.com/xray/latest/devguide/scorekeep-workerthreads.html. 

The reason why we use reflection is because we should not take a dependency on the X-RAY SDK in the batch module. When X-RAY is not available, this will lead to a no-op.

We also update the documentation with better guidance on when to choose which concurrency model.

Note for reviewers: Created separate issue for addressing code duplication: https://github.com/aws-powertools/powertools-lambda-java/issues/2302. 

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/1671

---

### DIFF
--- a/docs/utilities/batch.md
+++ b/docs/utilities/batch.md
@@ -484,7 +484,9 @@ used with SQS FIFO. In that case, an `UnsupportedOperationException` is thrown.
     in most cases the defaults work well, and changing them is more likely to decrease performance 
     (see [here](https://www.baeldung.com/java-when-to-use-parallel-stream#fork-join-framework)
     and [here](https://dzone.com/articles/be-aware-of-forkjoinpoolcommonpool)). 
-    In situations where this may be useful - such as performing IO-bound work in parallel - make sure to measure before and after!
+    In situations where this may be useful, such as performing IO-bound work in parallel, make sure to measure before and after!
+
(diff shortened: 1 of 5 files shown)


phipag:
@humanzz Would love your opinion here (looking at [your comment](https://github.com/aws-powertools/powertools-lambda-java/pull/1620#issuecomment-2083456666)).

Issue #2327, opened 2025-12-09
Bug: Metric dimensions are persisting across Lambda invocations


surajkumar:
### Expected Behaviour

When adding custom dimensions to metrics, they should **not** persist across Lambda invocations. 

### Current Behaviour

- Metric dimensions are persisting across Lambda invocations which causes inaccurate metric data. 
- This has been tested on version 2.1.0 and the latest 2.8.0

### Code snippet

```java
import com.amazonaws.services.lambda.runtime.Context;
import com.amazonaws.services.lambda.runtime.RequestHandler;
import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;
import software.amazon.lambda.powertools.logging.CorrelationIdPaths;
import software.amazon.lambda.powertools.logging.Logging;
import software.amazon.lambda.powertools.metrics.FlushMetrics;
import software.amazon.lambda.powertools.metrics.Metrics;
import software.amazon.lambda.powertools.metrics.MetricsFactory;
import software.amazon.lambda.powertools.metrics.model.DimensionSet;

import java.util.Map;

public class LambdaHandler implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
    private static final Logger LOGGER = LogManager.getLogger(LambdaHandler.class);
    private static final Metrics METRICS = MetricsFactory.getMetricsInstance();

    private boolean shouldEmitMetric = true;

    public LambdaHandler() {}

    @Override
    @Logging(correlationIdPath = CorrelationIdPaths.API_GATEWAY_REST, clearState = true)
    @FlushMetrics(namespace = "test-aws-stack", captureColdStart = true)
    public APIGatewayProxyResponseEvent handleRequest(APIGatewayProxyRequestEvent input, Context context) {
        LOGGER.info("Entry LambdaHandler#handleRequest");

        try {
            if (shouldEmitMetric) {
                Map<String, String> dimensions = Map.of("EXAMPLE_KEY", "EXAMPLE_VALUE");
                DimensionSet dimensionSet = new DimensionSet();
                dimensions.forEach(dimensionSet::addDimension);
                METRICS.addDimension(dimensionSet);
                METRICS.addMetric("SERL", 1d);

                shouldEmitMetric = false;
[...]

PR #2328, opened 2025-12-10
fix(metrics): Clear custom dimensions when flushing.

phipag:
## Summary

This PR clears custom dimensions when `.flush()` is called on the metrics instance. Note: This does not affect calls to `.flushMetrics()` since they create a temporary second metrics instance that supports flushing on its own with its own dimension set (copy). 

This was the default behavior in v1 of Powertools Java but is also the same behavior in the other runtimes such as TypeScript (see [comment](https://github.com/aws-powertools/powertools-lambda-java/issues/2327#issuecomment-3636551148)).

On flush, we clear:
- Custom dimensions and preserve default dimensions
- Metadata (also consistent with TS runtime)

### Changes

**Issue number:** https://github.com/aws-powertools/powertools-lambda-java/issues/2327

---

### DIFF
--- a/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
+++ b/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/internal/EmfMetricsLogger.java
@@ -164,6 +164,27 @@ public class EmfMetricsLogger implements Metrics {
             }
         } else {
             emfLogger.flush();
+
+            // Clear custom dimensions after flush while preserving default dimensions
+            clearCustomDimensions();
+        }
+    }
+
+    private void clearCustomDimensions() {
+        // Reset all dimensions in the EMF logger
+        emfLogger.resetDimensions(false);
+
+        // Re-apply default dimensions if they exist
+        if (!defaultDimensions.isEmpty()) {
+            DimensionSet emfDimensionSet = new DimensionSet();
+            defaultDimensions.forEach((key, value) -> {
+                try {
+                    emfDimensionSet.addDimension(key, value);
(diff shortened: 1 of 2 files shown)


phipag:
@surajkumar Thanks for testing! can you try to call as last statement in your testing code `METRICS.flush()`?

The reason why you see this warning is because `@FlushMetrics` uses a request-scoped proxy to a thread-local metrics instance (https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java#L33). Hence, `@FlushMetrics` emitted this warning because it did not collect any metrics. It created an instance which was not used because you manually created a separate EmfMetricsLogger.

surajkumar:
> @surajkumar Thanks for testing! can you try to call as last statement in your testing code `METRICS.flush()`?
> 
> The reason why you see this warning is because `@FlushMetrics` uses a request-scoped proxy to a thread-local metrics instance (https://github.com/aws-powertools/powertools-lambda-java/blob/main/powertools-metrics/src/main/java/software/amazon/lambda/powertools/metrics/MetricsFactory.java#L33). Hence, `@FlushMetrics` emitted this warning because it did not collect any metrics. It created an instance which was not used because you manually created a separate EmfMetricsLogger.

I tried with the flush at the end of the method and yes that works. 

I then re-tried with `MetricsFactory.getMetricsInstance()` (again, copying it over and updating where needed) and it gives me the same result of "No metrics were emitted" - Going to pull the repo, build and test - I'll get back to you.

phipag:
Awesome, thanks for confirming @surajkumar!

Merged this PR.

Issue #2004, opened 2025-08-04
Feature request: Priming for `powertools-tracing`


phipag:
### Use case

**PLEASE READ: Priming documentation**: https://github.com/aws-powertools/powertools-lambda-java/blob/main/Priming.md

Parent issue: https://github.com/aws-powertools/powertools-lambda-java/issues/1588
Sample PR: https://github.com/aws-powertools/powertools-lambda-java/pull/1861

Java [CRaC](https://openjdk.org/projects/crac/) can be used to prime an application by implementing `beforeCheckpoint()` and `afterRestore()` hooks in selected classes. When used with AWS Snapstart, the `beforeCheckpoint()` hook runs before the memory snapshot is taken. This behavior can be leveraged to further reduce restore durations by pre-loading classes and calling commonly used code to incorporate this into the memory snapshot.

1. Class-preloading / automatic priming: Based on a statically generated `classesloaded.txt` file, loads classes used at runtime into memory
2. Invoke priming: Execute commonly used code e.g. to initialize all reflectively access objects, warm up caches, TCP connection pools etc. Common examples include performing "dry" AWS SDK calls (not impacting production resources) or performing JSON serialization/deserialization. 

The goal of this issue is to implementing priming techniques for the `powertools-tracing` module.

### Solution/User Experience

Implement priming based on [Priming Documentation](https://github.com/aws-powertools/powertools-lambda-java/blob/main/Priming.md) (feel free to update documentation and suggest improvements):
- [ ] Automatic priming
- [ ] Invoke priming

### Alternative solutions

```Markdown

```

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (Java) Tenets](https://docs.powertools.aws.dev/lambda/java/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dcabib:
Hi @phipag! 👋

I'd be happy to help implement priming support for the `powertools-tracing` module. I've reviewed the priming documentation and the reference implementation in `powertools-metrics`, and I have a good understanding of what needs to be done.

__My proposed implementation approach:__

[...]

PR #2345, opened 2025-12-22
feat(crac): Support CRaC and priming of powertools tracing

Attyuttam:
### Changes

Added CRaC support in powertools-tracing, following the guide and blogs mentioned on the ticket.
Added automatic priming for powertools-metrics by preloading all the required classes, the list of classes is fetched from the JVM at compile time.

**Issue number: https://github.com/aws-powertools/powertools-lambda-java/issues/2004** 

---

### DIFF
--- a/powertools-tracing/pom.xml
+++ b/powertools-tracing/pom.xml
@@ -173,6 +173,13 @@
         </profile>
         <profile>
             <id>generate-graalvm-files</id>
+            <dependencies>
+                <dependency>
+                    <groupId>org.mockito</groupId>
+                    <artifactId>mockito-subclass</artifactId>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
             <build>
                 <plugins>
                     <plugin>
@@ -192,6 +199,13 @@
         </profile>
         <profile>
             <id>graalvm-native</id>
+            <dependencies>
+                <dependency>
+                    <groupId>org.mockito</groupId>
+                    <artifactId>mockito-subclass</artifactId>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
             <build>
                 <plugins>
                     <plugin>


phipag:
Hey @Attyuttam,

do you think you are able to develop inside Windows Subsytem for Linux (https://code.visualstudio.com/docs/remote/wsl) or inside a Docker container (https://code.visualstudio.com/docs/devcontainers/containers) or on a linux remote machine (https://code.visualstudio.com/docs/remote/remote-overview)?

As a quick solution, you could also make the changes locally and then build and run the tests inside a Docker container with Java and Maven installed. 

What do you think?

A large amount of changes you made in this PR are related to your OS (Windows) and break our Linux-based CI. Making all tests OS-independent would be task on its own.

Attyuttam:
sure @phipag , I will test as you suggested. Thanks !

Attyuttam:
Hi @phipag , I’ve updated the changes and validated them in a Linux environment using WSL to match the project’s CI setup.
Please let me know if any further adjustments are needed.

pom.xml
phipag, line 1: I think the changes in this file are not needed. In your previous PR I added this env var to all modules that need it.

Attyuttam: Sure

Attyuttam:
Thanks for the comments @phipag. I have incorporated them.

phipag:
Hey @Attyuttam,

this looks good now. Thanks! One last thing: You can see that the CI for GraalVM native is failing. The reason for this is that the Mockito inline mock maker is not supported for native unit tests and it fails now since you added mockito as a dependency. 

To fix this, can you please add the subclass mock maker as a dependency for both GraalVM maven profiles in the `pom.xml`? Here is an example for this from the metrics module:

https://github.com/aws-powertools/powertools-lambda-java/blob/54fe0be69938dbecb0d4efaa6c0152ee20c34eac/powertools-metrics/pom.xml#L179-L185

https://github.com/aws-powertools/powertools-lambda-java/blob/54fe0be69938dbecb0d4efaa6c0152ee20c34eac/powertools-metrics/pom.xml#L153-L159

phipag:
Awesome work @Attyuttam, I just tested this and it works. Thanks for your contribution! I am looking forward to your future PRs on the issues assigned to you. As always, let me know if you have any questions. 🥳

PR #2560, opened 2026-07-03
fix(common): strip leading whitespace in ClassPreLoader

### LINKED ISSUE #2559: Bug: SnapStart class priming loads 0 classes when classesloaded.txt has leading whitespace
## Expected Behaviour

`ClassPreLoader.preloadClasses()` loads every class listed in a module's `classesloaded.txt` during the SnapStart `beforeCheckpoint` phase, so those classes are baked into the snapshot and restore time drops.

## Current Behaviour

`ClassPreLoader` loads 0 classes from `powertools-tracing`, because every line in its `classesloaded.txt` starts with a leading space. The parser calls `line.stripTrailing()`, which removes trailing whitespace but keeps the leading space. `Class.forName(" java.lang.Object", ...)` then throws `ClassNotFoundException`, and the loader silently swallows it for all 5996 entries. The priming hook runs but primes nothing.

`powertools-tracing` ships this format today, so its automatic priming is a no-op in released versions. `powertools-large-messages` (PR #2533) uses the same format. The other modules (kafka, lambda-metadata, metrics, validation) have no leading space and are unaffected.

## Code snippet

```java
// ClassPreLoader.preloadClassesFromStream
final String className = line.stripTrailing(); // keeps the leading space
if (!className.isBlank()) {
    loadClassIfFound(className); // Class.forName(" java.lang.Object") -> ClassNotFoundException
}
```

## Possible Solution

Replace `line.stripTrailing()` with `line.strip()` so leading whitespace is removed as well. `strip()` tolerates both file formats, so no resource files need to change. I verified the fix on a deployed function: loaded classes went from 0 to 4062.

## Steps to Reproduce

1. Deploy a Lambda that uses `powertools-tracing` with SnapStart enabled (`ApplyOn: PublishedVersions`), and reference `TracingUtils` during INIT so the priming hook runs.
2. Add a temporary debug log in `ClassPreLoader` reporting how many classes loaded.
3. Observe that 0 classes load, because every `classesloaded.txt` entry carries a leading space.

## Powertools for AWS Lambda (Java) version

latest

## AWS Lambda function runtime

Java 11

## Debugging logs

```java
SnapStart priming: starting class preloading from 'classesloaded.txt'
SnapStart priming: successfully loaded 0 class(es) from resource
```

phipag:
### Changes

`ClassPreLoader` parsed `classesloaded.txt` with `line.stripTrailing()`, which removes trailing whitespace but keeps leading whitespace. Modules that ship the file with a leading space on every line, such as `powertools-tracing`, passed names like `" java.lang.Object"` to `Class.forName`. That call throws `ClassNotFoundException`, and the loader swallows it, so SnapStart class priming loaded 0 classes. `powertools-tracing` ships this format today, so its automatic priming is a no-op in released versions.

This change replaces `stripTrailing()` with `strip()`, which removes leading and trailing whitespace. `strip()` handles both file formats, so no resource files need to change.

I added a regression test that references a class with a leading space in the test `classesloaded.txt`. The test fails on the current code and passes with the fix. I also confirmed the fix on a deployed Lambda: preloaded classes went from 0 to 4062.

**Issue number:** #2559

---

### DIFF
--- a/powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
+++ b/powertools-common/src/main/java/software/amazon/lambda/powertools/common/internal/ClassPreLoader.java
@@ -21,12 +21,25 @@ import java.nio.charset.StandardCharsets;
 import java.net.URL;
 import java.net.URLConnection;
 import java.util.Enumeration;
+import java.util.regex.Pattern;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Used to preload classes to support automatic priming for SnapStart
  */
 public final class ClassPreLoader {
     public static final String CLASSES_FILE = "classesloaded.txt";
+
+    private static final Logger LOG = LoggerFactory.getLogger(ClassPreLoader.class);
+
+    // A binary class name is a series of dot-separated Java identifiers ($ is allowed for nested
+    // classes). This filters out malformed entries such as runtime-synthetic lambda classes
+    // (e.g. "com.example.Foo$$Lambda$1/0x0000...") and lines that contain a path or other junk,
+    // none of which Class.forName can load.
+    private static final Pattern BINARY_CLASS_NAME = Pattern.compile(
+            "[\\p{L}_$][\\p{L}\\p{N}_$]*(\\.[\\p{L}_$][\\p{L}\\p{N}_$]*)*");
 
     private ClassPreLoader() {
         // Hide default constructor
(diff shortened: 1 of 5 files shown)


phipag:
The PMD `CloseResource` finding on `ClassPreLoader.java:59` is a false positive on pre-existing code, because the stream is closed by the `try (is; ...)` block in `preloadClassesFromStream`.

hjgraca (approved):

